PR #8, opened 2024-05-05
feat: add eslint plugins (import, react)

43081j:
Adds docs for replacements of `eslint-plugin-import` and `eslint-plugin-react`.


43081j:
The solution here is to fix the plugin 👍

Let's continue tracking in the issue you linked. It will certainly get fixed so we have no action to do here I think

Issue #9, opened 2024-05-12
`eslint-plugin-n` instead of `eslint-plugin-node`


Lehoczky:
eslint-plugin-node is no longer maintained and has been forked as eslint-plugin-n.

Their repositories: 
- https://github.com/mysticatea/eslint-plugin-node
- https://github.com/eslint-community/eslint-plugin-n

43081j:
would be good to add that to the docs here:
https://github.com/es-tooling/module-replacements/tree/main/docs/modules

and then we can pull it into eslint-plugin-depend too

Lehoczky:
No worries, I will do it tomorrow ;)

PR #10, opened 2024-05-13
feat: add `eslint-plugin-node`

Lehoczky:
Related: #9

### DIFF
--- a/docs/modules/README.md
+++ b/docs/modules/README.md
@@ -13,6 +13,7 @@ ESLint plugin.
 ## List of modules
 
 - [`eslint-plugin-import`](./eslint-plugin-import.md)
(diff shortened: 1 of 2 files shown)


43081j (approved):
looks good to me

thanks very much for the contribution 🙏

43081j:
if you get time, you could also add it to the eslint plugin here:

https://github.com/es-tooling/eslint-plugin-depend/blob/969f80d6005a1872b7ea3a41e12fa181c76e6328/src/replacements.ts#L33-L59

adding another entry to the array with the right path should be enough (`moduleName` is the npm module you're making redundant, and `docPath` is the name of the doc you just added)

Lehoczky:
Consider it done 🫡

PR #26, opened 2024-06-24
feat: add replacement for deep-equal

benmccann:


docs/modules/deep-equal.md
```diff
@@ -0,0 +1,13 @@
+# deep-equal
+
+`deep-equal` has [roughly 50 dependencies](https://npmgraph.js.org/?q=deep-equal) while alternatives have zero.
```
43081j, line 3: i would reword this to be less specific about the number of dependencies i think (as it may change)

maybe just something like `deep-equal has various lighter weight and faster alternatives`

benmccann: thanks. good point. I've updated it

I also confirmed that saying faster is correct: https://www.npmjs.com/package/fast-deep-equal#performance-benchmark

43081j (approved):

benmccann:
Good thought. It's pretty good being zero dependencies. Though it is noticably larger than the other two on bundlephobia, so maybe I'll leave it off for now unless you still think we should add it

43081j:
no worries, works for me 👍

Issue #27, opened 2024-06-24
replacements for custom promise implementations


dominikg:
q has been marked as deprecated 2 weeks ago: https://github.com/kriskowal/q/issues/866#issuecomment-2161615101 still has 10k dependents.

bluebird even has >30k https://github.com/petkaantonov/bluebird

bluebird is a bit more tricky as they have api that isn't covered by native promises yet but a reimplementation based on them should be considerably smaller.

PR #28, opened 2024-06-24
feat: add bluebird replacement

benmccann:
ref https://github.com/es-tooling/module-replacements/issues/27


benmccann:
makes sense! I've moved it there

manifests/native.json
```diff
@@ -215,6 +215,14 @@
...
-    }
+    },
+    {
+      "type": "native",
+      "moduleName": "bluebird",
+      "nodeVersion": "0.12.0",
+      "replacement": "Promise",
+      "mdnPath": "Global_Objects/Promise",
+      "category": "native"
+    },
```
valadaptive, line 226: This trailing comma is what's causing CI to fail

benmccann: ah, thanks! I hadn't noticed it failing

43081j (approved):

manifests/native.json
```diff
@@ -215,6 +215,14 @@
       "replacement": "Array.prototype.lastIndexOf",
       "mdnPath": "Global_Objects/Array/lastIndexOf",
       "category": "native"
+    },
+    {
+      "type": "native",
+      "moduleName": "bluebird",
+      "nodeVersion": "0.12.0",
+      "replacement": "Promise",
```
zloirock, line 223: `bluebird` also contains many helpers, so it would be better to create a page and explain it.

43081j: fair point 🤔 i thought it belongs as a `native` replacement since we can define when `Promise` became available etc

but maybe i made a mistake here and it belongs in `preferred` after all?

what do you think?

zloirock: Since in this case is possible to create a page with an explanation, I think that `preferred` is a better place.

43081j: @benmccann would you mind opening a new PR with your original change, and undo this one?

i didn't realise bluebird has various other exports

benmccann: done here: https://github.com/es-tooling/module-replacements/pull/51

PR #36, opened 2024-06-26
feat: add replacement for tempy

valadaptive:
Not sure what Node version to put this at. `fs.mkdtemp` itself goes back to Node v5, but `fs.rmdir(..., {recursive: true}` is Node 12.x and up, and it now gives you a big deprecation message telling you to use `fs.rm` instead, which is Node 14.x and up.

It shouldn't be too much of a problem, however, since `tempy` is also Node 14.x and up.

### DIFF
--- a/docs/modules/tempy.md
+++ b/docs/modules/tempy.md
@@ -1,0 +1,42 @@
+# tempy
+
+`tempy` can often be replaced with built-in Node APIs.
+
+# Alternatives
+
+## NodeJS (since v14.x)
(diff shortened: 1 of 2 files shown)


43081j:
was going to suggest this belongs in  the `native` manifest rather than `preferred`, but i think the fact you need extra code rather than replacing a single expression probably means you're right here

since at least `mkdtemp` is available pretty much everywhere, think we're all good 👍

PR #38, opened 2024-06-26
Add `has`

zloirock:
Here is a funny discussion https://github.com/tarruda/has/pull/17

Now the initial package creator has fixed it, however, it's time to use `Object.hasOwn`.


43081j (approved):
good catch, looks good to me!

PR #41, opened 2024-06-26
feat: replacements for mkdirp and rimraf

valadaptive:
These packages' functionality have been integrated into Node's `fs` module, and can usually be removed.

### DIFF
--- a/docs/modules/mkdirp.md
+++ b/docs/modules/mkdirp.md
@@ -1,0 +1,19 @@
+# mkdirp
+
+`mkdirp` can often be replaced with built-in Node APIs.
+
+# Alternatives
+
+## NodeJS (since v10.x)
+
+Node.js v10.0 and up supports the `recursive` option in [the `fs.mkdir`
+function](https://nodejs.org/api/fs.html#fspromisesmkdirpath-options), which
+allows parent directories to be created automatically:
+
+```js
+import {mkdir} from 'node:fs/promises';
+import {mkdirSync} from 'node:fs';
+
+await mkdir('some/nested/path', {recursive: true});
+mkdirSync('some/nested/path', {recursive: true});
+```
--- a/docs/modules/rimraf.md
+++ b/docs/modules/rimraf.md
@@ -35,3 +35,11 @@ async function main() {
 rmdirSync(path, {recursive: true});
 
 ```
+
+For command-line usage, such as npm scripts, you can run Node in eval mode:
+```bash
+node -e "fs.rmSync('./foo', { recursive: true, force: true })"
+```
+Note that this may not work for developers who use alternative JavaScript
+runtimes and do not have Node installed. To support them, you can use the
+[premove](https://www.npmjs.com/package/premove) package instead.
--- a/docs/modules/rimraf.md
+++ b/docs/modules/rimraf.md
@@ -12,16 +12,16 @@ function](https://nodejs.org/api/fs.html#fspromisesrmpath-options), which allows
 files and directories to be deleted recursively:
 
 ```js
-import {rm} from 'node:fs/promises';
-import {rmSync} from 'node:fs';
+import { rm } from 'node:fs/promises';
+import { rmSync } from 'node:fs';
 
 // This will throw an error if the path does not exist.
-await rm(path, {recursive: true});
-rmSync(path, {recursive: true});
+await rm(path, { recursive: true });
+rmSync(path, { recursive: true });
 
 // This will do nothing if the path does not exist.
-await rm(path, {recursive: true, force: true});
-rmSync(path, {recursive: true, force: true});
+await rm(path, { recursive: true, force: true });
+rmSync(path, { recursive: true, force: true });
 ```
 
 If you need to support Node 12, you can [use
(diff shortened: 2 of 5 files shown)


valadaptive:
This could be put in `native` if there's some way to attach documentation--I feel like at the moment, `native` can't really be used for anything more complex than polyfills where all you have to do is swap out the name of the function.

benmccann:
perhaps we should also update this? lol

https://github.com/es-tooling/module-replacements/blob/218ebd1658d60e2e572b92f8408bed8300b7f49b/package.json#L39

we could also show how to replace it when it's used in this way

valadaptive:
> perhaps we should also update this? lol
> 
> https://github.com/es-tooling/module-replacements/blob/218ebd1658d60e2e572b92f8408bed8300b7f49b/package.json#L39
> 
> we could also show how to replace it when it's used in this way

Good catch! I've added a section at the bottom for replacing rimraf in a CLI context, and gone ahead and done so in this repo.

docs/modules/rimraf.md
```diff
@@ -0,0 +1,45 @@
...
+    await rmdir(path, {recursive: true});
+}
+
+rmdirSync(path, {recursive: true});
+
+```
+
+For command-line usage, such as npm scripts, you can run Node in eval mode:
+```bash
+node -e "fs.rmSync('./foo', { recursive: true, force: true })"
```
benmccann, line 41: Here we have `force` but not above. Is that right that we need it in this case and not the others?

benmccann: I just looked up the docs as I wasn't actually sure what `force` did. It says:

> When true, exceptions will be ignored if path does not exist. Default: false

Maybe worth removing it here, but adding a more general note about it

valadaptive: `force` will ignore errors if the path doesn't exist. I've added that to the documentation.

docs/modules/rimraf.md
```diff
@@ -0,0 +1,50 @@
...
+```js
+const {rmdir} = require('fs').promises;
+const {rmdirSync} = require('fs');
+
+async function main() {
+    await rmdir(path, {recursive: true});
+}
+
+rmdirSync(path, {recursive: true});
+
```
benmccann, line 41: most projects use more spacing in their code. I wonder if we should setup prettier markdown to standardize this
```suggestion
import { rm } from 'node:fs/promises';
import { rmSync } from 'node:fs';

// This will throw an error if the path does not exist.
await rm(path, { recursive: true });
rmSync(path, { recursive: true });

// This will do nothing if the path does not exist.
await rm(path, { recursive: true, force: true });
rmSync(path, { recursive: true, force: true });
```

If you need to support Node 12, you can [use
`fs.rmdir`](https://nodejs.org/api/fs.html#fspromisesrmdirpath-options), which
also has the `recursive` option available since Node v12.10, although it will
print a deprecation message in Node v14 and up:

```js
const { rmdir } = require('fs').promises;
const { rmdirSync } = require('fs');

async function main() {
    await rmdir(path, { recursive: true });
}

rmdirSync(path, { recursive: true });
```

valadaptive: Fixed the bracket spacing

docs/modules/mkdirp.md
```diff
@@ -0,0 +1,19 @@
+# mkdirp
+
+`mkdirp` can often be replaced with built-in Node APIs.
```
43081j, line 3: does `mkdirp` _only_ export `mkdir` and `mkdirSync`? i.e. it has no other exported functionality/utils?

if so, we should probably add it to the native manifest and specify the correct node version there

docs/modules/rimraf.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+For command-line usage, such as npm scripts, you can run Node in eval mode:
+```bash
+node -e "fs.rmSync('./foo', { recursive: true, force: true })"
+```
+Note that this may not work for developers who use alternative JavaScript
+runtimes and do not have Node installed. To support them, you can use the
+[premove](https://www.npmjs.com/package/premove) package instead.
```
43081j, line 50: can you move this into its own section?

while you can use `node -e` on the command line, i can see it still being useful to have a purpose built CLI for this

so we should list `premove` as one of the alternatives

docs/modules/mkdirp.md
valadaptive: Modulo a couple minor differences (you can pass in a custom `fs` implementation to `mkdirp`).

As I mentioned above:
> This could be put in `native` if there's some way to attach documentation--I feel like at the moment, `native` can't really be used for anything more complex than polyfills where all you have to do is swap out the name of the function.

I see the `native` list as only for polyfills where the function behaves identically to the built-in one. If you need to pass in an extra argument, as is the case here, there should be some way to document that requirement.

Also, functions in the `native` manifest need to have an MDN path, which this does not, as it's not a JavaScript builtin but instead part of Node's standard library.

43081j: ah sorry i missed that!

lets put it in `native` but keep the doc, and we can figure out how to provide doc links some other time

43081j: actually, completely ignore me 😆 

sorry! lets keep it here as im really not sure how we should best handle node-specific natives yet

docs/modules/rimraf.md
valadaptive: I've moved `premove` to its own section.

AbhiPrasad (commented):
Many people also rely on the glob functionality with `rimraf`. I think we should make that clear, and show you can use `fs.glob` (introduced with Node 22) in conjunction with `fs.rm` to have that behaviour.

valadaptive:
`fs.glob` is still experimental; I'd prefer to wait a bit before I'm comfortable recommending it.

43081j (approved):

43081j:
i think its true that rimraf users will be using globs sometimes, but also true they probably don't need to (and shouldn't be).

i think lets stick with these docs for now, but see if any issues are raised re having a glob alternative. if it is more important than we thought, we can document using glob, fdir or fs.glob alongside

PR #45, opened 2024-06-27
Add some too bloated `String` polyfills

zloirock:
Resume #39

Without adding polyfills for recent features.

I'm not sure where should be added https://github.com/es-shims/es-string-html-methods since it contains polyfills for some features and breaks the `native` schema.

### DIFF
--- a/manifests/native.json
+++ b/manifests/native.json
@@ -258,6 +258,118 @@
     },
     {
       "type": "native",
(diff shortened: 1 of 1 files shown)


43081j (approved):

43081j:
i think we would leave proposals/recent features etc out of the manifests until they exist natively

so looks good to me as is 👍

PR #67, opened 2024-06-30
Add `jQuery`

### LINKED ISSUE #13: jQuery alternatives
The exact alternatives could be:

- native JS
- some more robust frameworks, like Angular, React, Vue
- some lighter alternatives, like [htmx ](https://htmx.org/)or [Alpine.js](https://alpinejs.dev/)

zloirock:
Fixes #13.

### DIFF
--- a/docs/modules/jquery.md
+++ b/docs/modules/jquery.md
@@ -1,0 +1,17 @@
+# jQuery
+
+jQuery is a great project that at one time greatly simplified the life of web developers. However, now almost all features from this project can be easily replaced with native JavaScript methods so it makes little sense to load a dependency this large to provide them.
+
+# Replacements
+
+## You might not need jQuery
+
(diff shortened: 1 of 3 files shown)


docs/modules/jquery.md
```diff
@@ -0,0 +1,17 @@
+# jQuery
+
+jQuery is a great project that at one time greatly simplified the life of web developers. However, now almost all features from this project can be easily replaced with native JavaScript methods and it makes no sense to load this quite a big dependence.
```
43081j, line 3: i would reword the last bit as:

> so it makes little sense to load a dependency this large to provide them

seems like `dependence` should be `dependency` at least

zloirock: Ok. Updated.

43081j (approved):

PR #69, opened 2024-07-01
fix: Add missing entries to module docs

AbhiPrasad:
Adding missing entries for `mkdirp`, `rimraf`, and `sort-object`.

I also added a basic validation script so we don't forgot to keep `docs/modules/README.md` updated! This was done in this commit: 270163372a84f6df52389494c46bb306d6228aa3. If the validation stuff feels too heavy-handed, lmk and I can revert it!


package.json
```diff
@@ -4,14 +4,16 @@
...
-    "format": "prettier --write \"./src/**/*.ts\" \"./manifests/**/*.json\"",
+    "build": "npm run clean && npm run build:types",
     "build:types": "tsc --noEmit",
     "build:update-manifest-paths": "node scripts/update-manifests-dist-path.js",
-    "build": "npm run clean && npm run build:types",
+    "clean": "node -e \"fs.rmSync('./dist', {force: true, recursive: true})\"",
+    "format": "prettier --write \"./src/**/*.ts\" \"./manifests/**/*.json\"",
+    "generate-schema": "node scripts/generate-schema.js",
+    "prepare": "tshy && npm run build:update-manifest-paths",
+    "test:validate": "npm run validate-manifests && npm run validate-module-list",
```
AbhiPrasad, line 14: I alpha ordered the scripts, but lmk and I can not do this.

43081j (approved):
good idea. i've been wanting to add something to validate this for a while 👍

Issue #58, opened 2024-06-29
`resolve`


AbhiPrasad:
We can replace [`resolve`](https://www.npmjs.com/package/resolve) with built-ins starting Node `8.9.0` and above.

It [pulls in a bunch of deps](https://npmgraph.js.org/?q=resolve), and doesn’t support exports field and custom conditions (which is very common in modern JS apps).

<img width="756" alt="image" src="https://github.com/es-tooling/module-replacements/assets/18689448/13dbca16-0379-4aa7-be50-d670ae19215a">

For CJS you can use `require.resolve`. ESM is more tricky, you prob need https://www.npmjs.com/package/resolve.exports, but there is [`import.meta.resolve`](https://nodejs.org/api/esm.html#importmetaresolvespecifier) that was introduced with Node `20.6.0`

~Here's an example PR I opened up with rollup that does the ESM thing~: https://github.com/rollup/plugins/pull/1742 (I may have to change this to make it work with ESM, need more testing)

Here's an example PR (that tbh should be merged in to `require-in-the-middle`) that does the CJS port: https://github.com/elastic/require-in-the-middle/pull/25

AbhiPrasad:
From @antfu on [discord](https://discord.com/channels/1227627367204257852/1227653370945474590/1256413814132314122):

> Resolve is a very complex problem to solve. `require.resolve` only resolve CJS conditions. For ESM and custom conditions, you might want https://www.npmjs.com/package/resolve.exports, but that still, wouldn't even be closer to a 1:1 replacement of resolve.

AbhiPrasad:
https://packages.ecosyste.ms/registries/npmjs.org/packages/resolve/dependent_packages?kind=runtime&order=desc&sort=downloads

benmccann:
I don't think we can just suggest this one to be replaced. Sometimes you can, but very often projects are using APIs from `resolve` that are not available in `require.resolve`

ljharb:
fwiw you can replace `supports-preserve-symlinks-flag` with `export default typeof window === 'undefined' ? true : null;` if it's node 6.2+ (and if you have separate node and browser files, one is just `true` and the other just `null`.

I think `is-core-module` you're already replacing, and `path-parse` can be replaced with `export { parse } from 'path'` in node >= 0.12.

Sadly `import.meta.resolve` is both flagged and broken in all sorts of ways, so it's likely not yet a viable option.

AbhiPrasad:
[...]

PR #71, opened 2024-07-01
feat: Add path-parse to micro-utilities manifest

AbhiPrasad:
ref: https://github.com/es-tooling/module-replacements/issues/58#issuecomment-2197869098

https://www.npmjs.com/package/path-parse

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -77,6 +77,12 @@
       "moduleName": "is-even",
       "replacement": "Use (n % 2) === 0",
       "category": "micro-utilities"
+    },
+    {
+      "type": "simple",
+      "moduleName": "path-parse",
+      "replacement": "Use built-in parse via import { parse } from 'node:path'",
+      "category": "micro-utilities"
     }
   ]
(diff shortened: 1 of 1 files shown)


AbhiPrasad:
Actually nvm this should be in `native.json` - sorry didn't realize that first. 

@43081j I'll close this, but how should we approach these node polyfills?

43081j:
i think we need to come up with a way to have node-specific natives

we can add them to the native manifest, but they wouldn't have an MDN path of course. so they'd probably end up being a `simple` replacement like you have here

i wonder if we need a new type at some point, like `native-node` or `native-runtime`

AbhiPrasad:
So I spent some time thinking about this, and I think a new `native-node` type makes the most sense to me, because these directly apply to Node.js APIs. 

This also means that `native` means isomorphic runtime agnostic replacements, and we can do `native-X` for runtime specific stuff (for example `native-browser`). I can see a world where we also have `native-deno` or similar too.

If that feels reasonable, I can open a PR - and we can discuss further there?

Issue #72, opened 2024-07-01
`is-core-module`


AbhiPrasad:
https://www.npmjs.com/package/is-core-module

ref: https://github.com/es-tooling/module-replacements/issues/58#issuecomment-2197869098

You can do something like so:

```ts
import module from 'node:module';

function isCoreModule(moduleName): boolean {
  if (moduleName.startsWith('node:') {
    return true;
  }

  // requires v18.6.0, v16.17.0
  if (module.isBuiltin(moduleName)) {
    return true;
  }

  // if using Node.js older than 16.17
  // use: module.builtinModules.includes(mod)

  return false;
}
```

We should prob add https://www.npmjs.com/package/is-builtin-module and https://github.com/sindresorhus/builtin-modules to the list as well (although those packages do make it pretty clear in the README).

43081j:
we already have `is-builtin-module` here:
https://github.com/es-tooling/module-replacements/blob/main/docs/modules/is-builtin-module.md?plain=1

if this provides the same functionality, we should probably just add an entry to the `preferred` manifest pointing at the same doc

AbhiPrasad:
I can't believe I missed `is-builtin-module` :sob:

Let me make a PR!

PR #79, opened 2024-07-02
feat: Add is-core-module replacement

AbhiPrasad:
resolves https://github.com/es-tooling/module-replacements/issues/72

`is-core-module` can be replaced the same way `is-builtin-module` can.


docs/modules/is-builtin-module.md
AbhiPrasad, line 1: Should I rename the file? Debating this.

43081j: think its fine as is 👍 

there will probably be other packages which do this too so we inevitably have to choose one name

Issue #80, opened 2024-07-02
`aggregate-error`


AbhiPrasad:
https://github.com/sindresorhus/aggregate-error

From the README

> Note: With [Node.js 15](https://medium.com/@nodejs/node-js-v15-0-0-is-here-deb00750f278), there's now a built-in [AggregateError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError) type.

Removing this gets rid of 4 deps / 23 KB, and anyone supporting Node.js 16 or higher should not be using this.

PR #81, opened 2024-07-02
feat: Add replacement for aggregate-error / es-aggregate-error

AbhiPrasad:
resolves https://github.com/es-tooling/module-replacements/issues/80

Adds entry for `aggregate-error`, and documents `es-aggregate-error` which was already in the `native.json` manifest.

### DIFF
--- a/docs/modules/README.md
+++ b/docs/modules/README.md
@@ -12,9 +12,11 @@ ESLint plugin.
 
 ## List of modules
 
+- [`aggregate-error`](./aggregate-error.md)
 - [`bluebird`](./bluebird.md)
 - [`cpx`](./cpx.md)
 - [`deep-equal`](./deep-equal.md)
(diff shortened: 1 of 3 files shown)


docs/modules/README.md
```diff
@@ -12,9 +12,11 @@ ESLint plugin.
 
 ## List of modules
 
+- [`aggregate-error`](./aggregate-error.md)
 - [`bluebird`](./bluebird.md)
 - [`cpx`](./cpx.md)
 - [`deep-equal`](./deep-equal.md)
+- [`es-aggregate-error`](./es-aggregate-error.md)
```
zloirock, line 19: Since it's a polyfill, I'm not sure that it requires its own docs page.

43081j: it may be that we shouldn't add these to the docs 🤔 

since its a native feature with an MDN doc

what do you think?

PR #82, opened 2024-07-02
Add `call-bind`

zloirock:


43081j (approved):

manifests/micro-utilities.json
```diff
@@ -1,5 +1,11 @@
 {
   "moduleReplacements": [
+    {
+      "type": "simple",
+      "moduleName": "call-bind",
+      "replacement": "Use Function.call.bind(v)",
+      "category": "micro-utilities"
+    },
```
ljharb, line 8: this is only true for the default export; `call-bind/callBound` is a bit more complex.

PR #87, opened 2024-07-03
Add `define-data-property` and `define-accessor-property`

zloirock:


manifests/micro-utilities.json
```diff
@@ -6,6 +6,12 @@
       "replacement": "Use Function.call.bind(v)",
       "category": "micro-utilities"
     },
+    {
+      "type": "simple",
+      "moduleName": "define-data-property",
+      "replacement": "Use Object.defineProperty(object, key, { value, enumerable, configurable, writable })",
```
zloirock, line 12: Maybe, it should be in the `native` manifest. WDYT?

43081j: Since it is a one to one replacement, I think so

If it was more ambiguous, it would belong here

zloirock: This util works only with the creation of data properties, `Object.defineProperty` also can be used to change descriptors and create accessors.

zloirock: But yes, `Object.defineProperty` is one to one replacement, just with another signature.

zloirock: Updated.

43081j (approved):

PR #91, opened 2024-07-04
Add more replacements: make-dir, object-keys, object-assign, readable-stream

### LINKED ISSUE #37: `mkdirp`
https://stackoverflow.com/questions/37589161/is-it-possible-to-create-a-dir-in-node-like-mkdir-p-does

### LINKED ISSUE #62: `make-dir`
`fsPromises.mkdir('…', {recursive: true})`

onigoetz:
* `make-dir` ( Fixes #62 ) ( Fixes #37 is already fixed )
* `object.keys` with `Object.keys` (this is identical to the `object-keys` package
* `object-assign` with `Object.assign` (this is identical to the `object.assign` package
* `readable-stream` that can be replaced by `node:stream`

### DIFF
--- a/docs/modules/readable-stream.md
+++ b/docs/modules/readable-stream.md
@@ -1,0 +1,9 @@
+# readable-stream
+
+[`readable-stream`](https://www.npmjs.com/package/readable-stream) is a mirror of the NodeJS `stream` module.
+
+# Alternatives
+
+## NodeJS (since v0.9.4)
+
+If your NodeJS version is recent it is better to directly use [`stream`](https://nodejs.org/api/stream.html).
--- a/docs/modules/readable-stream.md
+++ b/docs/modules/readable-stream.md
@@ -1,9 +1,17 @@
 # readable-stream
 
-[`readable-stream`](https://www.npmjs.com/package/readable-stream) is a mirror of the NodeJS `stream` module.
+[`readable-stream`](https://www.npmjs.com/package/readable-stream) is a mirror of the NodeJS `stream` module which also works in the browser.
 
 # Alternatives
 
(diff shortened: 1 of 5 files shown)


docs/modules/readable-stream.md
```diff
@@ -0,0 +1,9 @@
+# readable-stream
+
+[`readable-stream`](https://www.npmjs.com/package/readable-stream) is a mirror of the NodeJS `stream` module.
+
+# Alternatives
+
+## NodeJS (since v0.9.4)
+
+If your NodeJS version is recent it is better to directly use [`stream`](https://nodejs.org/api/stream.html).
```
zloirock, line 9: Maybe makes sense to mention streams from web standards, like https://developer.mozilla.org/en-US/docs/Web/API/Streams_API ?

onigoetz: It could, but they aren't a direct alternative to NodeJS streams, they need to be converted from one to the other, no?

Still, I can add another alternative

zloirock: Maybe. I just don't know is `readable-stream` NodeJS-only package or not. If yes - it's not required, if not - I think that it should be mentioned.

zloirock: > v4.x.x of readable-stream is a cut from Node 18. This version supports Node 12, 14, 16 and 18, **as well as evergreen browsers**

onigoetz: ah yes indeed, I added an alternative for Streams API

zloirock (approved):

43081j (approved):

PR #93, opened 2024-07-04
Add `q`

zloirock:


docs/modules/README.md
```diff
@@ -12,7 +12,7 @@ ESLint plugin.
 
 ## List of modules
 
-- [`bluebird`](./bluebird.md)
+- [`bluebird` / `q`](./bluebird-q.md)
```
43081j, line 15: do you think there are many other promise libraries we will add?

if its just these two i think this is fine. but if it'll grow, we should probably just name it after one of them and word it something like `bluebird and similar libraries`

zloirock: Most likely, but now there are only 2, these are specific libraries, and they have recommendations to move to native `Promise` in their docs. When it will be extended, it will be renamed as it's required.

43081j: Makes sense to me 👍

43081j (approved):

PR #116, opened 2024-07-23
feat: add globbing section

SuperchupuDev:


docs/modules/globby.md
```diff
@@ -0,0 +1,32 @@
+# globby
+
+`globby` is a globbing library that presents [some behaviors](https://github.com/sindresorhus/globby#features)
+no other libraries have, mainly the handling of negated patterns and directory expansion.
+
+It pulls in [20+](https://npmgraph.js.org/?q=globby) transitive dependencies and can be replaced
```
43081j, line 6: I wouldn't mention the number of dependencies as it can change. its better to just mention there are lighter alternatives (that's what we do in other docs too)

docs/modules/globby.md
```diff
@@ -0,0 +1,32 @@
...
+with lighter and faster alternatives.
+
+# Alternatives
+
+## tinyglobby
+
+[tinyglobby](https://www.npmjs.com/package/tinyglobby) can be used as an alternative in most cases.
```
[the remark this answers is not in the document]
SuperchupuDev: done. i mentioned npm in lowercase though as it's incorrect to use it in all caps. could send a pr later changing the template to have the proper mention

image

docs/modules/globby.md
```diff
@@ -0,0 +1,46 @@
...
+
+`globby` is a globbing library that presents [some behaviors](https://github.com/sindresorhus/globby#features)
+no other libraries have, mainly the handling of negated patterns and directory expansion. It can be replaced
+with lighter and faster alternatives.
+
+# Alternatives
+
+## tinyglobby
+
+`tinyglobby` can be used as an alternative in most cases.
```
43081j, line 11: i would reword this as something like:

> `tinyglobby` is a smaller alternative. It does not have the same options as globby but often can be a drop-in replacement.

docs/modules/globby.md
```diff
@@ -0,0 +1,46 @@
...
+
+[Project Page](https://github.com/SuperchupuDev/tinyglobby)
+
+[npm](https://npmjs.com/package/tinyglobby)
+
+## fdir + picomatch
+
+If you don't rely on `globby`'s special features, you can just use
+`fdir` combined with `picomatch`.
+Its API is nothing like `globby`'s, but it does the job if configured properly.
```
43081j, line 22: similar, here i would reword something like:

> fdir is a fast alternative which can optionally support globs via picomatch. It has a different API shape to globby but can achieve the same goals

docs/modules/globby.md
```diff
@@ -0,0 +1,46 @@
...
+
+[`fdir` Project Page](https://github.com/thecodrr/fdir)
+
+[`fdir` on npm](https://npmjs.com/package/fdir)
+
+[`picomatch` Project Page](https://github.com/micromatch/picomatch)
```
43081j, line 28: i wouldn't bother linking to picomatch since the fdir docs should mention it already in their section on globbing

docs/modules/globby.md
```diff
@@ -0,0 +1,46 @@
...
+
+[`picomatch` on npm](https://npmjs.com/package/picomatch)
+
+## Node.js (since v22.x)
+
+Node.js v22 and up supports the [`fs.glob`](https://nodejs.org/api/fs.html#fspromisesglobpattern-options)
+function. Although it's (currently) not as fast as other alternatives, it works as a zero-dependency
+solution. It doesn't have `globby`'s special features though.
```
43081j, line 36: i don't think we need the last sentence about globby features

docs/modules/globby.md
```diff
@@ -0,0 +1,46 @@
...
+
+## fast-glob
+
+`fast-glob` does pretty much the same as `globby`
```
43081j, line 40: do you think we should just leave this out if it is bloated?

fdir can cover the same use cases pretty much

43081j (commented):
sorry for another round of comments but just had a few points on the wording

let me know if it makes sense

docs/modules/globby.md
SuperchupuDev: i feel like that would be misleading as it does have the same options and api, its just that it doesn't cover all of them yet

docs/modules/globby.md
SuperchupuDev: how else would readers know about compatibility aspects between the two? some could think it covers all globby features

docs/modules/globby.md
43081j: > It doesn't include all `globby` options though.

is the part i feel could be worded better

we're more concerned with the common use case than whatever edge case options globby might support. if tinyglobby is a faster alternative, we should make that clear. if it is lighter, that too. ultimately these libs are used mostly just to glob a simple pattern and we don't need to focus on what options are available

docs/modules/globby.md
43081j: because what exactly are globby's "special features"? from this text, that isn't clear. and if you make it clear, you're going to basically duplicate globby's readme

what is the common use case of globby? if it is to glob patterns, regardless of whatever niche options they have, we probably don't need to care so much about the options

SuperchupuDev: it's not about its options but rather about its default behavior, like autoconverting i.e a `src` pattern to a `src/**/*` one. i'd say that's the whole point of globby - globbing with added behavior that can be considered useful. users of libraries that blindly switch to alternatives without taking that into account will end up seeing breaking behavior if those patterns are user provided (in fact that's what happened in tsup after me forgetting to add some globby behavior before it was suggested to move that logic into tinyglobby)

43081j: if some globby-specific feature is that important, we should probably be defining exactly which feature is missing in each of these replacements

nobody knows what globby's "special features" are. we need to be exact here and say specifically what isn't supported

SuperchupuDev: thought the lines all the way to the top were clear enough, will do 👍

docs/modules/glob.md
```diff
@@ -0,0 +1,37 @@
+# Globbing
+
+Most globbing libraries are bloated and have lighter and faster alternatives.
```
43081j, line 3: nit: i would reword throughout as `glob libraries`

also this should probably say `Many glob libraries` rather than `Most` (since we will be reducing it over time, so it won't be true anymore)

docs/modules/glob.md
```diff
@@ -0,0 +1,37 @@
...
+even need to use globs with this package. It can be combined with `picomatch` if you do need them.
+
+[Project Page](https://github.com/thecodrr/fdir)
+
+[npm](https://npmjs.com/package/fdir)
+
+## Node.js (since v22.x)
+
+Node.js v22 and up supports the [`fs.glob`](https://nodejs.org/api/fs.html#fspromisesglobpattern-options)
+function. Although it's (currently) experimental, it works as a zero-dependency solution.
```
43081j, line 20: i probably wouldn't mention that it is experimental as that'll change over time

i think it'll be enough to just mention it exists as long as we're clear about the version constraint

docs/modules/glob.md
```diff
@@ -0,0 +1,37 @@
...
+
+`tinyglobby` can be used as an alternative to `globby` and `fast-glob`. It uses `fdir` and `picomatch`
+under the hood.
+
+`fast-glob` is a library that presents a custom behavior of negated patterns no other glob libraries have.
+
+`globby` wraps around it to add even more [default behaviors](https://github.com/sindresorhus/globby#features)
+that are not straightforward to replace.
+
+`tinyglobby` should only be needed if you rely on these default behaviors.
```
43081j, line 32: you can word this less like a last resort i think. something like:

> `tinyglobby` is a good replacement if you rely on these behaviours, or you want to continue using a similar API to existing glob libraries

43081j (approved):

PR #118, opened 2024-07-26
feat: Add manifests to exports map

rschristian:
Adds the manifests to the export map so they can be accessed by tools that utilize Node's module resolution. For example, if you tried to run this:

```js
import nativeReplacements from 'module-replacements/manifests/native.json' with { type: 'json' };

console.log(nativeReplacements);
```

You'd see this:

```
node:internal/modules/esm/resolve:299
  return new ERR_PACKAGE_PATH_NOT_EXPORTED(
         ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './manifests/native.json' is not defined by "exports" in /home/ryun/Projects/deptree/node_modules/module-replacements/package.json imported from /home/ryun/Projects/deptree/index.js
```

Also adds `package.json` to the exports map too -- it's a handy thing that many other libraries offer. Never know, someone might want to access it in a script for some reason.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,11 @@
         "types": "./dist/commonjs/main.d.ts",
         "default": "./dist/commonjs/main.js"
       }
-    }
+    },
+    "./package.json": "./package.json",
+    "./manifests/micro-utilities.json": "./manifests/micro-utilities.json",
+    "./manifests/native.json": "./manifests/native.json",
+    "./manifests/preferred.json": "./manifests/preferred.json"
   },
   "types": "./dist/commonjs/main.d.ts",
   "type": "module"


43081j (approved):

43081j:
makes sense to me 👍

rschristian:
@43081j It looks like this got stripped out of the `2.3.0` publish, I'm assuming this isn't intentional?

https://www.runpkg.com/?module-replacements@2.3.0/package.json

43081j:
good catch. this is because tshy generates the export map for us, so we should've actually added it to the tshy config rather than directly to the map

have opened #131 to fix

rschristian:
Gotcha. Sorry, was not/am not familiar with tshy, I assumed I could just add it directly to the export map (did it from the web interface IIRC, didn't even pull locally)

PR #120, opened 2024-07-27
feat: Add replacement for classnames -> clsx

andriyor:


docs/modules/README.md
```diff
@@ -34,3 +34,4 @@ ESLint plugin.
 - [`sort-object`](./sort-object.md)
 - [`tempy`](./tempy.md)
 - [`uri-js`](./uri-js.md)
+- [`classnames`](./classnames.md)
```
43081j, line 37: can you keep the list sorted? (a-z)

43081j (approved):
looks good to me

although `classnames` doesn't seem so problematic. it is pretty small as far as i can see

would be good to get some more opinions on this

43081j:
after discussing this on discord, im not sure if we should add it

classnames doesn't seem like a problematic dependency, even if there are some faster alternatives. it isn't bloated from what i can see, and isn't crazily slow or unmaintained

ill close this for now, but if you can explain why you think it should be replaced, maybe we can reopen

if its just a preference (i.e. there's nothing wrong with `classnames` but clsx is faster), it may be that we reopen this once we add a new, more opinionated manifest

PR #125, opened 2024-07-29
chore: update typo

bluwy:
I believe this part means older projects that aren’t actively maintained (?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ This is a very opinionated list of modules with preferred replacements. Often
 these replacements are much lighter or more modern than the modules they are
 replacing.
 
-Sometimes these may also be forks of older packages which are actively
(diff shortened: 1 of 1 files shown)


bluwy:
Ah I see the confusion now. No I think I understand that `these` refer to the replacements too. I think it comes to this:

When I read the sentence, I assume ` which are actively
maintained (unlike the source module)` refers to the `older packages`, not `these`. So I'm confused why older packages are still actively maintained.

 Maybe we should add a comma before the `which`? Or I think if others read the way I read too, this PR also works, because it implies `these` to be more actively maintained.

43081j:
> Sometimes these may also be actively maintained forks of older, unmaintained source packages.

maybe this will do better?

43081j (approved):

PR #127, opened 2024-08-01
feat: add missing packages

thepassle:
I noticed there were some discrepancies between the codemods we have, and the packages that are in module-replacements. I think the codemods should always follow what's in module-replacements, so here's a PR that adds the missing packages. Once this is merged and released I'll add a script in the codemods repo that checks on PR to see if the contributed codemod is actually in module-replacements or not to avoid discrepancies in the future.

### DIFF
--- a/docs/modules/md5.md
+++ b/docs/modules/md5.md
@@ -1,5 +1,7 @@
 # md5
 
+If you're using the `md5` package, you should consider using a more secure algorithm if possible, if not you can find a native Node.js alternative in the example blow.
+
 # Alternatives
 
 ## NodeJS
--- a/docs/modules/md5.md
+++ b/docs/modules/md5.md
@@ -1,0 +1,11 @@
+# md5
+
+# Alternatives
+
+## NodeJS
+
+```js
+import crypto from 'node:crypto';
+
+crypto.createHash('md5').update('message').digest('hex')`
+```
--- a/manifests/native.json
+++ b/manifests/native.json
@@ -1039,6 +1039,94 @@
       "replacement": "Array.prototype.lastIndexOf",
       "mdnPath": "Global_Objects/Array/lastIndexOf",
       "category": "native"
+    },
+    {
+      "type": "native",
+      "moduleName": "abort-controller",
+      "nodeVersion": "0.10.0",
+      "replacement": "AbortController",
+      "mdnPath": "Global_Objects/AbortController",
+      "category": "native"
+    },
+    {
+      "type": "native",
+      "moduleName": "array.prototype.findlast",
+      "nodeVersion": "0.10.0",
+      "replacement": "Array.prototype.findLast",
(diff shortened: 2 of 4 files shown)


manifests/native.json
```diff
@@ -1039,6 +1039,94 @@
...
+      "nodeVersion": "0.10.0",
+      "replacement": "String.prototype.repeat",
+      "mdnPath": "Global_Objects/String/repeat",
+      "category": "native"
+    },
+    {
+      "type": "native",
+      "moduleName": "is-builtin-module",
+      "nodeVersion": "0.10.0",
+      "replacement": "Use `isBuiltin` from 'node:module'",
```
thepassle, line 1111: This one and `md5` are native, but don't have an mdnPath, how do you want to handle those? Just leave it empty?

43081j: it should be a `simple` entry, or a `documented` one if we need extra wording around migration

in this case, we already have it here:

https://github.com/es-tooling/module-replacements/blob/7ab526fdcc36e2be4a09d2b28e2e0c16aa606b0d/manifests/preferred.json#L51-L62

manifests/native.json
```diff
@@ -1039,6 +1039,94 @@
...
+      "moduleName": "abort-controller",
+      "nodeVersion": "0.10.0",
+      "replacement": "AbortController",
+      "mdnPath": "Global_Objects/AbortController",
+      "category": "native"
+    },
+    {
+      "type": "native",
+      "moduleName": "array.prototype.findlast",
+      "nodeVersion": "0.10.0",
```
zloirock, line 1054: The Node version is incorrect for most of those packages. However, I agree that they should be replaced everywhere.

manifests/native.json
43081j: for the `md5` one, we should probably just add it to the preferred manifest and give it a doc

what do you think?

manifests/native.json
thepassle: good catch :) lazy copy/paste work on my side, will fix

thepassle: fixed

manifests/native.json
thepassle: moved md5 to preferred, removed is-builtin-module; it was already there

thepassle:
@43081j PTAL

docs/modules/md5.md
```diff
@@ -0,0 +1,11 @@
+# md5
+
```
43081j, line 2: can you add a small description? just mention that you should use a more secure algorithm if possible, but otherwise can use built-in functionality

43081j (approved):

docs/modules/md5.md
thepassle: done

43081j (approved):

PR #129, opened 2024-08-01
fix is-number example

talentlessguy:


43081j:
indeed, which is what my second example would achieve: `typeof n === 'number' || !isNaN(n)`

the extra logic is checking if the number is `Infinity` and returning `false` if it is

i'm saying we probably don't need to do any of that because it isn't the common use case

talentlessguy:
I think it's best not to ask to replace with typeof, in case somebody's relying on string parsing

43081j:
that is what `|| !isNaN(n)` would achieve

maybe im not being clear enough, let me try explain again:

the code you've added here (and what `is-number` does) is doing two things:

- `n - n === 0` is basically checking it isn't `Infinity` and isn't `NaN`
- `v.trim() !== ""` is accounting for the fact `+""` is `0`

what i'm saying is we shouldn't be emulating what `is-number` does. we should be pushing people to use simpler code

`typeof n === 'number' || !isNaN(n)` will allow literal numbers and number-like strings, without the extra cruft around checking for infinity and what not

recommend what you want to see used, rather than what is currently used

talentlessguy:
ah alright, sorry 

now I get it

I'll update my PR accordingly

Issue #115, opened 2024-07-23
add `neoqs` to `qs`'s doc


SuperchupuDev:

SuperchupuDev:
apparently done in #140 a long time ago

PR #140, opened 2024-08-07
feat: add `neoqs`

Jayllyz:
#115

### DIFF
--- a/docs/modules/qs.md
+++ b/docs/modules/qs.md
@@ -17,12 +17,14 @@
 [picoquery](https://www.npmjs.com/package/picoquery) can be used when you need nesting and arrays.
 
 ## qs-esm
+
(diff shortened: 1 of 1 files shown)


PuruVJ:
Haha thanks for this!

docs/modules/qs.md
```diff
@@ -21,3 +21,11 @@
 
 - No polyfills for legacy Node versions, making this package dependency-free and reducing bundle-size.
 - Includes TypeScript types, eliminating the need to install a separate @types/ package.
+
+## neoqs
+[neoqs](https://www.npmjs.com/package/neoqs) is a fork of `qs` with the following core differences:
```
43081j, line 26: nit: can you add a new line after the heading and do the same for `qs-esm` (seems we missed it)

43081j:
the qs doc is actually inconsistent and hasn't followed the template every other doc follows

i'll open an issue and we can sort that separately 👍 

this looks good though

Issue #142, opened 2024-08-07
qs doc is inconsistent


43081j:
The documentation for `qs` doesn't follow the standard doc template

We should update it to follow the same structure as all the other docs. It is currently missing the footer links

43081j:
Fixed by #143

PR #143, opened 2024-08-07
fix: `qs` inconsistent documentation

Jayllyz:
#142


43081j (approved):
looks good! thanks for picking it up so quickly 🥳

PR #146, opened 2024-08-10
Publish to codemod.com registry

mohebifar:
This PR adds a script that auto publishes codemods to the codemod.com registry. It also adds a README file for all codemods that will accompany the published codemod in the registry.

### DIFF
--- a/codemods/abort-controller/README.md
+++ b/codemods/abort-controller/README.md
@@ -1,0 +1,21 @@
+# Abort Controller Codemod
+
+## Introduction
+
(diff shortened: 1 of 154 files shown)


43081j:
Should this be in the codemods repo instead? 

cc @thepassle to decide if we should do this too

thepassle:
Yeah I think this PR went to the wrong repo

mohebifar:
Yep, sorry, I forked the wrong repo. 🤦

PR #150, opened 2024-08-20
feat: add alternatives to process execution libraries

### LINKED ISSUE #141: `ezspawn` and `@jsdevtools/ez-spawn`
can be migrated to `tinyexec`

Namchee:
## Overview

Closes #141

This pull request adds alternative replacements for process execution libraries such as `execa` and `zx`.


43081j:
although zx is fairly big, i think we should leave it out as its a very well maintained/stable project for when you want templated scripts

note to anyone else here: before this can land we need to introduce a new manifest I think too. since execa is useful in some situations (where you heavily rely on its extended features). so it may be that we introduce a new opinionated manifest or something that can somehow conditionally be suggested based on usage

good contribution 🎉

Namchee:
Thanks for the review. `zx` dropped.

Do you have other alternatives that can be used? I'm thinking of [`nanoexec`](https://github.com/fabiospampinato/nanoexec) but I'm not sure about it.

As for `ezspawn` (and friends), I don't think it's worth it to include them here considering they're barely used.

docs/modules/execa.md
```diff
@@ -0,0 +1,23 @@
+# Execa
+
+Execa is process execution library that simplifies command execution by providing programmatic API on top of `node:child_process`.
```
43081j, line 3: this isn't really consistent with the rest of the repo. we don't need to explain what `execa` does, rather we just need to mention why you should move off it

in this case, you'd move off it because there are lighter alternatives when you don't need extended features

otherwise we're just duplicating info you can get from execa's own readme

Namchee:
I disagree with `zx` as an alternative as the unpacked size is actually bigger than `execa` but doesn't provide compelling features that `execa` doesn't have (cmiiw).

The only benefit of `zx` is that it doesn't contain any dependencies.

[npmgraph for `execa`](https://npmgraph.js.org/?q=execa#select=exact%3Aexeca%409.3.1)
[npmgraph for `zx`](https://npmgraph.js.org/?q=zx#select=exact%3Azx%408.1.4)

docs/modules/execa.md
Namchee: Valid point, should I remove the list of important features too? Personally, I want to add it since users might not know what those 'kitchen-sink' features are.

43081j: i would just mention that if you need templating or other high level features, you might not want to switch. soon as we start listing any features though, it'll lead us down a rabbit hole and we'll end up replicating execa's readme again

remember its the job of the manifest/tooling to suggest this replacement under the right circumstances. once we have a usage-based lint rule and codemod, this'll only be suggested when the user doesn't use any of those extended features

so we don't need to go into much depth here

43081j:
fair point. we shouldn't have it in the replacements list, though, as it is one of the better choices for when you need templating

we can work on reducing the size of zx meanwhile, at least

so i would remove it from the list, and from the replacements list for now

docs/modules/execa.md
```diff
@@ -0,0 +1,13 @@
+# Execa
+
+When you don't need [additional features](https://github.com/sindresorhus/execa/#features) that `execa` provides such as [string templating](https://github.com/sindresorhus/execa/blob/main/docs/execution.md#template-string-syntax), [inter-process communication](https://github.com/sindresorhus/execa/#exchange-messages), and [debugging features](https://github.com/sindresorhus/execa/?tab=readme-ov-file#detailed-error),  you can replace it with tinier alternatives below.
```
43081j, line 3: > execa and other execution libraries have lighter alternatives available, especially if you do not need a higher level abstraction (e.g. templating syntax, scripting, etc).

something as simple as this would be fine as this doc won't always be about execa

we don't need to explain why someone should use execa here. this doc is here to explain why one would move off execa and other libraries.

it also needs renaming i think to `process-exec.md` (same as your links in the `README`)

docs/modules/README.md
```diff
@@ -41,3 +42,4 @@ ESLint plugin.
 - [`tempy`](./tempy.md)
 - [`traverse`](./traverse.md)
 - [`uri-js`](./uri-js.md)
+- [`zx`](./process-exec.md)
```
43081j, line 45: this one can go away i think

docs/modules/execa.md
```diff
@@ -0,0 +1,13 @@
...
+
+# Alternatives
```
43081j, line 5: we can possibly add:

- [zurk](https://github.com/webpod/zurk)
- [nanoexec](https://github.com/fabiospampinato/nanoexec)

let me know what you think

docs/modules/README.md
Namchee: Whoops forgot to delete that.

docs/modules/execa.md
Namchee: From what I've read from `zurk` is that it's a form of experiment for `zx` and presumably be removed once `zx` has moved to v8. Is it OK for us to include temporary stuff like `zurk`.

As for `nanoexec`, I'll be adding that.

docs/modules/execa.md
Namchee: Rewritten based on your feedbacks

docs/modules/execa.md
43081j: if thats true im happy to leave zurk out until zx figures out what they're doing with it 👍

43081j (approved):

manifests/preferred.json
```diff
@@ -54,6 +54,12 @@
       "docPath": "eslint-plugin-react",
       "category": "preferred"
     },
+    {
+      "type": "documented",
+      "moduleName": "execa",
+      "docPath": "execa",
```
43081j, line 60: the build failure is because this should be `process-exec` now

PR #161, opened 2024-09-28
docs: add body-parser

### LINKED ISSUE #160: bodyparser -> milliparsec

Namchee:
## Overview

Closes #160.

This pull request adds section for `body-parser` and its alternative (namely, `milliparsec`)


docs/modules/README.md
```diff
@@ -14,6 +14,8 @@ ESLint plugin.
 
 - [`@jsdevtools/ez-spawn`](./process-exec.md)
 - [`bluebird` / `q`](./bluebird-q.md)
+- [`body-parser`](./body-parser.md)
+- [`builtin-modules`](./is-builtin-module.md)
```
43081j, line 18: Did you mean to add this line?

Was it missing before?

Namchee: Yes, they're surprisingly not available before.

43081j: ah it exists u nder `is-builtin-module` here:

https://github.com/es-tooling/module-replacements/blob/7d864c64f68394ea55f33ef67c712c217e5e68c4/docs/modules/README.md?plain=1#L30

Namchee: Oops, I missed that since I thought the entries were lexicographically sorted by the markdown name. Removed.

43081j (approved):
looks good to me, but lets get at least one other opinion before we merge 👍

docs/modules/body-parser.md
```diff
@@ -0,0 +1,12 @@
+# Body-parser
+
+`body-parser` can be replaced with more modern alternatives which is both lighter and faster.
```
SuperchupuDev, line 3:
```suggestion
`body-parser` can be replaced with more modern alternatives which are both lighter and faster.
```
small grammar mistake

TheThing:
Discussion on a discord server revealed that we should probably wait until default post limit is implemented into millisec. Don't want to give malicious people the ability to DDoS a server from this recommendation.

43081j:
It may also be worth explaining that some simple use cases don't need a package

Maybe we list "native" as an alternative and give some examples?

43081j (approved):

Issue #126, opened 2024-07-29
`fs-extra` native replacement


bluwy:
Most of the fs-extra APIs can be replaced with native fs calls these days especially on later node versions. Eg some migration I made: https://github.com/vitejs/vite/pull/17782

We probably needs a migration guide from fs-extra to fs mapping its APIs: https://github.com/jprichardson/node-fs-extra?tab=readme-ov-file#methods. However, I’ve not found a solid guide yet elsewhere, except for https://ar.al/2021/03/07/fs-extra-to-fs/ but it’s 3 years ago.

43081j:
we could try document it in the replacement doc. we haven't done that so far but fs-extra does need a bit of a migration guide i think

we can still follow the doc template but have an extra section

walter-0:
I'd like to start adding the applicable fs-extra modules into the native manifest

bluwy:
@ziebam has a lot of experience migrating `fs-extra` to `fs` and could help review the changes

bluwy:
fixed in #173

PR #173, opened 2024-10-16
Feat: Add `fs-extra`

walter-0:
Closes (#126)

Snippets to add:

- [ ] `emptyDir`
- [ ] `emptyDirSync`
- [ ] `ensureFile`
- [ ] `ensureFileSync`
- [ ] `ensureLink`
- [ ] `ensureLinkSync`
- [ ] `ensureSymlink`
- [ ] `ensureSymlinkSync`
- [ ] `outputFile`
- [ ] `outputFileSync`
- [ ] `outputJson`
- [ ] `outputJsonSync`
- [ ] `pathExists`
- [ ] `readJson`
- [ ] `readJsonSync`
- [ ] `writeJson`
- [ ] `writeJsonSync`

### DIFF
--- a/docs/modules/fs-extra.md
+++ b/docs/modules/fs-extra.md
@@ -20,7 +20,7 @@ import * as fs from 'node:fs/promises';
 To use the callback and sync APIs:
 
 ```js
-import * as fs from 'node:fs/promises';
+import * as fs from 'node:fs';
 ```
 
 ### Equivalent Methods
--- a/docs/modules/fs-extra.md
+++ b/docs/modules/fs-extra.md
@@ -40,37 +40,3 @@ import * as fs from 'node:fs';
 #### Methods that need migrating or more than one `node:fs` method
 
 For all other methods, a combination of fs methods may be needed rather than there being a one-to-one replacement. See the docs [here](https://nodejs.org/docs/latest/api/fs.html)
-
-##### `emptyDir(dir[, callback])`
-
-##### `emptyDirSync(dir)`
-
-##### `ensureFile(file[, callback])`
-
-##### `ensureFileSync(file)`
-
-##### `ensureLink(srcPath, destPath[, callback])`
-
-##### `ensureLinkSync(srcPath, destPath)`
-
-##### `ensureSymlink(srcPath, destPath[, type][, callback])`
-
-##### `ensureSymlinkSync(srcPath, destPath[, type])`
-
-##### `outputFile(file, data[, options][, callback])`
-
-##### `outputFileSync(file, data[, options])`
-
-##### `outputJson(file, object[, options][, callback])`
-
-##### `outputJsonSync(file, object[, options])`
-
-##### `pathExists(file[, callback])`
-
-##### `readJson(file[, options][, callback])`
-
-##### `readJsonSync(file[, options])`
-
-##### `writeJson(file, object[, options][, callback])`
-
-##### `writeJsonSync(file, object[, options])`
--- a/docs/modules/fs-extra.md
+++ b/docs/modules/fs-extra.md
@@ -23,38 +23,52 @@ To use the callback and sync APIs:
 import * as fs from 'node:fs';
 ```
 
-### Equivalent Methods
+#### Equivalent Methods
 
-- `ensureFile` -> `fs.exists` - deprecated since v1
-- `ensureFileSync` -> `fs.existsSync`
-- `copySync` -> `fs.copyFileSync`
-- `mkdirp` -> fs.mkdir`
-- `mkdirpSync` -> `fs.mkdirSync`
-- `removeSync` -> `fs.rmSync`
-- `remove` -> `fs.rm`
-- `move` -> `fs.rename`
-- `moveSync` -> `fs.renameSync`
-- `ensureDirSync` -> `fs.existsSync`
-- `pathExistsSync` -> `fs.existsSync`
-- `outputFileSync` -> `fs.writeFileSync`
+| fs-extra                                                                                                      | node:fs                                                   | Notes                                                                                                                                                                                                                                                                                                                                                                                                                                                                        |
+| ------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `copySync(src, dest[, options])`                                                                              | `fs.copyFileSync(src, dest[, mode])`                      |                                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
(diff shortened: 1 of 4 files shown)


docs/modules/fs-extra.md
```diff
@@ -0,0 +1,60 @@
...
+To use the promise-based APIs:
+
+```js
+import * as fs from 'node:fs/promises';
+```
+
+To use the callback and sync APIs:
+
+```js
+import * as fs from 'node:fs/promises';
```
ziebam, line 23:
```suggestion
import * as fs from 'node:fs';
```

docs/modules/fs-extra.md
```diff
@@ -0,0 +1,60 @@
...
+- `copySync` -> `fs.copyFileSync`
+- `mkdirp` -> fs.mkdir`
+- `mkdirpSync` -> `fs.mkdirSync`
+- `removeSync` -> `fs.rmSync`
+- `remove` -> `fs.rm`
+- `move` -> `fs.rename`
+- `moveSync` -> `fs.renameSync`
+- `ensureDirSync` -> `fs.existsSync`
+- `pathExistsSync` -> `fs.existsSync`
+- `outputFileSync` -> `fs.writeFileSync`
```
ziebam, line 39: It would probably be good to mention what options need to be passed to match `fs-extra`'s behavior, e.g. `recursive: true` for `mkdirp`. I can help with that if needed.

docs/modules/fs-extra.md
```diff
@@ -0,0 +1,60 @@
...
+
+To use the callback and sync APIs:
+
+```js
+import * as fs from 'node:fs/promises';
+```
+
+### Equivalent Methods
+
+- `ensureFile` -> `fs.exists` - deprecated since v1
```
ziebam, line 28: Since `fs.exists` is deprecated, I think we should suggest either using the synchronous version (`fs.existsSync`) or inlining what `fs-extra` does (and Node docs recommend): `fs.access().then(() => true).catch(() => false)`.

docs/modules/fs-extra.md
```diff
@@ -0,0 +1,60 @@
...
+- `ensureFileSync` -> `fs.existsSync`
```
ziebam, line 29: `ensureFileSync` creates a file if it doesn't exist, so just `fs.existsSync` doesn't suffice. I can help write snippets for such cases where one `fs-extra` function needs to be replaced with multiple `node:fs` ones.

docs/modules/fs-extra.md
```diff
@@ -0,0 +1,60 @@
...
+- `copySync` -> `fs.copyFileSync`
+- `mkdirp` -> fs.mkdir`
+- `mkdirpSync` -> `fs.mkdirSync`
+- `removeSync` -> `fs.rmSync`
+- `remove` -> `fs.rm`
+- `move` -> `fs.rename`
+- `moveSync` -> `fs.renameSync`
+- `ensureDirSync` -> `fs.existsSync`
```
ziebam, line 37: Same situation as with `ensureFileSync`, although this can be quite simply replaced with `fs.mkdir(dir, { recursive: true })` as it won't overwrite the directory if it exists.

docs/modules/fs-extra.md
ziebam, line 39: `outputFileSync` also creates the parent directory if it doesn't exist.

docs/modules/fs-extra.md
```diff
@@ -0,0 +1,60 @@
...
+- `pathExistsSync` -> `fs.existsSync`
+- `outputFileSync` -> `fs.writeFileSync`
+
+### Methods that need migrating
+
+- `copy`
```
ziebam, line 43: While the two work slightly differently under the hood, for most cases `copy` can be replaced with `node:fs`'s `cp` that's available since Node 16.

docs/modules/fs-extra.md
```diff
@@ -0,0 +1,60 @@
...
+- `emptyDir`
+- `emptyDirSync`
+- `ensureDir`
```
ziebam, line 46: `ensureDir` is an alias for `mkdirp`, so it can be replaced in the same way. There are generally many aliases in the `fs-extra` package that we can consider documenting here.

docs/modules/fs-extra.md
```diff
@@ -0,0 +1,60 @@
...
+- `ensureLink`
+- `ensureLinkSync`
+- `ensureSymlink`
+- `ensureSymlinkSync`
+- `mkdirs`
+- `mkdirsSync`
```
ziebam, line 52: Same with these, they are aliases for `mkdirp` and `mkdirpSync` respectively.

docs/modules/fs-extra.md
```diff
@@ -0,0 +1,60 @@
...
+- `outputFile`
+- `outputJson`
+- `outputJsonSync`
+- `pathExists`
+- `readJson`
+- `readJsonSync`
+- `writeJson`
+- `writeJsonSync`
```
ziebam, line 60: We can provide some snippets here that should be enough for most migrations. The ones I've done always use the defaults (e.g. no reviver function and two spaces), and it's straightforward to do that with `JSON.parse()` and `JSON.stringify()`.

walter-0:
Hey @ziebam thanks a lot for the comments! My intent for this PR was mostly to get `fs-extra` merged in so that I can start making a codemod that covers everything (since the module-replacement-codemods need to exist in this repo first), and not as a complete migration guide. If this is ok as an initial, then more detail/snippets can be added to this doc so I can work on a codemod concurrently.

43081j:
I'll have a full read through in the next day or so. But from a quick scan through it, I think we should do a few of the PR comments at least since they're about which method to prefer. 

I don't think we need a migration guide, but we should at least mention the basic mappings (old to new/replacement)

docs/modules/fs-extra.md
```diff
@@ -0,0 +1,74 @@
...
+| `pathExistsSync(file)`                                                                                        | `fs.existsSync(path)`                                     |                                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
+
+#### Methods that need migrating or more than one `node:fs` method
+
+##### `emptyDir(dir[, callback])`
```
43081j, line 42: for all of these, I would just mention something to summarise instead:

> For all other methods, a combination of `fs` methods may be needed rather than there being a one-to-one replacement. See the docs [here](#)

docs/modules/fs-extra.md
```diff
@@ -0,0 +1,76 @@
...
+| `pathExistsSync(file)`                                                                                        | `fs.existsSync(path)`                                     |                                                                                                                                                                                                                                                                                                                                                                                                                                                                              |
+
+#### Methods that need migrating or more than one `node:fs` method
+
+For all other methods, a combination of fs methods may be needed rather than there being a one-to-one replacement. See the docs [here](https://nodejs.org/docs/latest/api/fs.html)
+
+##### `emptyDir(dir[, callback])`
+
+##### `emptyDirSync(dir)`
```
43081j, line 46: if you can remove these empty headings I think we're good and can get this merged 👍

43081j (approved):

Issue #152, opened 2024-08-21
dotenv


Jayllyz:
Suggest dotenvx instead

Mouvedia:
You mean [`dotenvx`](https://dotenvx.com).

jonasb:
What about Node's built-in `--env-file` https://nodejs.org/dist/latest-v20.x/docs/api/cli.html#--env-fileconfig? I prefer it since it's one less dependency, at least for simple use cases. (tsx, bun and deno also support `--env-file`)

43081j:
Good find! Had no idea node has this built in now

If we don't already, we should mention it in the replacements repo for dotenv

PR #177, opened 2024-11-10
Mention nodejs `--env-file-if-exists` in dotenv alternatives

kouak:
This relatively new nodejs flag is even closer in behavior to `import 'dotenv/config'`, since node will not refuse to start if the env-file does not exist.

Ref #152


43081j (approved):
Makes sense to me! 

Thank you

PR #178, opened 2024-11-11
feat: add `grapheme-splitter` and `grapheme` alternatives

cometkim:
Discussed in https://github.com/es-tooling/ecosystem-cleanup/issues/70

### DIFF
--- a/manifests/native.json
+++ b/manifests/native.json
@@ -1047,22 +1047,6 @@
       "replacement": "Object.assign, or if deep clones are needed, use structuredClone",
       "mdnPath": "Global_Objects/Object/assign",
       "category": "native"
-    },
-    {
-      "type": "native",
-      "moduleName": "grapheme-splitter",
-      "nodeVersion": "16.0.0",
-      "replacement": "Intl.Segmenter",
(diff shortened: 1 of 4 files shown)


manifests/native.json
```diff
@@ -1047,6 +1047,22 @@
       "replacement": "Object.assign, or if deep clones are needed, use structuredClone",
       "mdnPath": "Global_Objects/Object/assign",
       "category": "native"
+    },
+    {
```
43081j, line 1051: I think we should only have it in the preferred manifest. Otherwise we might cause duplicate warnings in linters etc

cometkim: Hmm ok. But then, I guess other modules will have similar issues, does it even make sense to maintain a native manifest?

43081j: so far, any given module has only been in one manifest

the native one makes sense for those modules which have an exact native equivalent. e.g. if we only suggested using `Intl` and not a third party package, it'd probably exist only in the native manifest

mentioning the third party package means it belongs in the preferred manifest with documentation

I think that's fine though, its good to mention the package too

PR #188, opened 2024-12-06
docs: add chalk to preferred manifest

### LINKED ISSUE #149: `cli-color`, `chalk`
In the short-term, replace with `picocolors`, `kleur`, etc.

Maybe starting in April 2025 when Node 18 is EOL consider replacing with `styleText`

43081j:
Adds `chalk` and `cli-color` to the preferred manifest and documents alternatives.

Closes #149

### DIFF
--- a/docs/modules/chalk.md
+++ b/docs/modules/chalk.md
@@ -1,0 +1,35 @@
+# chalk
+
+`chalk` and similar libraries have various alternatives which are lighter and faster.
+
+# Alternatives
+
+## picocolors
+
+A widely used alternative which is much lighter and faster.
+
+[Project Page](https://github.com/alexeyraspopov/picocolors)
+
+[npm](https://npmjs.com/package/picocolors)
+
+## ansis
+
+A lighter alternative which also supports 256 color creation.
+
+[Project Page](https://github.com/webdiscus/ansis)
+
+[npm](https://npmjs.com/package/ansis)
+
+## styleText (Node 20.x and above)
+
+Node itself has a `styleText` function in the built-in `util` library.
+
+You can use it like so:
+
+```ts
+import { styleText } from 'node:util';
+
+styleText('green', 'Success!')
+```
+
+[Project Page](https://nodejs.org/api/util.html#utilstyletextformat-text-options)
--- a/docs/modules/chalk.md
+++ b/docs/modules/chalk.md
@@ -1,6 +1,6 @@
 # chalk
 
-`chalk` and similar libraries have various alternatives which are lighter and faster.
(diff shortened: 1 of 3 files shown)


docs/modules/chalk.md
```diff
@@ -0,0 +1,35 @@
...
+
+[npm](https://npmjs.com/package/ansis)
+
+## styleText (Node 20.x and above)
+
+Node itself has a `styleText` function in the built-in `util` library.
+
+You can use it like so:
+
+```ts
```
SuperchupuDev, line 29: is it really necessary for it to be typescript? extremely small nit

```suggestion
```js
```

docs/modules/chalk.md
```diff
@@ -0,0 +1,35 @@
...
+
+## ansis
+
+A lighter alternative which also supports 256 color creation.
+
+[Project Page](https://github.com/webdiscus/ansis)
+
+[npm](https://npmjs.com/package/ansis)
+
+## styleText (Node 20.x and above)
```
SuperchupuDev, line 23: considering this is a stable feature that is in the two latest major lts versions, should it be moved up in the list? considering is the smallest (no deps) option :thinking:

SuperchupuDev (commented):
small nits

also you said all was welcome on the discord server, so i'm adding to the review my dad's cats as well
<details><summary>cats</summary>
<p>

image

</p>
</details>

docs/modules/chalk.md
43081j: It's a habit more than anything just in case I end up writing types. Happy to change it

docs/modules/chalk.md
43081j: The important one I think is LTS. If it is in all LTS versions then it's pretty much widely available

SuperchupuDev: 20.x and 22.x are both lts now :eyes: although 18.x is still supported, just in maintenance

43081j: Let's leave it for now and we can bump it up once 18 falls out of maintenance (apparently in a few months)

docs/modules/chalk.md
```diff
@@ -0,0 +1,35 @@
+# chalk
+
+`chalk` and similar libraries have various alternatives which are lighter and faster.
```
benmccann, line 3: we should probably mention `cli-color` here somewhere since we're pointing to this page for it. otherwise I think it could be a bit confusing
```suggestion
`chalk`, `cli-color`, and similar libraries have various alternatives which are lighter and faster.
```

PR #190, opened 2024-12-20
docs: mention `@stylistic/eslint-plugin-jsx` and `typescript` in `eslint-plugin-react` alternatives

Rel1cx:


docs/modules/eslint-plugin-react.md
```diff
@@ -6,7 +6,53 @@
...
+- `react/jsx-sort-default-props`
+- `react/jsx-sort-props`
+- `react/jsx-space-before-closing`
+- `react/jsx-tag-spacing`
+- `react/jsx-wrap-multilines`
+
+[Project Page](https://eslint.style/packages/jsx)
+[npm](https://www.npmjs.com/package/@stylistic/eslint-plugin-jsx)
+
+## TypeScript
```
43081j, line 42: im not sure if we should bother listing this typescript section. as we don't really explain what replaces them or how (and that doesn't belong in these docs. rather it'd be better once we start work on our extended docs)

Rel1cx: > im not sure if we should bother listing this typescript section. as we don't really explain what replaces them or how (and that doesn't belong in these docs. rather it'd be better once we start work on our extended docs)

Good point. I'll remove the TypeScript section for now.

PR #194, opened 2025-02-04
fix(fs-extra): add missing `recursive` option to `cp`

serhalp:
fs-extra's `copy` supports files and directories: https://github.com/jprichardson/node-fs-extra/blob/176ad6a6dbdec503c2d79fa70d17df32dff49972/docs/copy.md

`cp` from `node:fs` only supports files by default — it needs `recursive: true` to handle directories: https://nodejs.org/docs/latest-v18.x/api/fs.html#fscpsrc-dest-options-callback


43081j (approved):
makes sense to me 👍 

good catch

serhalp:
Ah, technically it should also include `verbatimSymlinks: true`... Rarely matters but it did just bite me. Worth adding?

43081j:
i think we can leave it. we don't need to document the exact equivalent, people just need to know `node:fs` can do it with whatever combination of options you prefer

PR #207, opened 2025-05-05
chore: register eslint-plugin-unicorn

Namchee:
## Overview

This pull request registers `eslint-plugin-unicorn` to preferred module section.

> Probably gonna delete this if the upstream is interested in merging the changes instead of keeping the fork alive.


43081j (approved):

43081j:
given it seems it'll be a long term fork, we can merge i think

i can't see a future where the source catches up. fisker is on board with most perf changes, but sindre isn't on board with any dependency change.

PR #208, opened 2025-05-09
Revert "docs: register eslint-plugin-unicorn (#207)"

TheDevMinerTV:
This reverts commit 6f4b8eefb8843658c68c10776f6e3cfd98104763 because of community backlash (also because @43081j asked me to).


43081j:
@Namchee I think we preemptively added this and it seems to have caused a little drama 😅

So I'm going to take it back out for now while the fork gains some more usage etc. If it gets to a point where it is popular and we can't contribute back upstream, we can reintroduce this

Issue #48, opened 2024-06-27
Add a custom lint rule to check for duplicates in the manifests


43081j:
We should add a simple eslint rule (and introduce eslint with that) which basically looks for duplicates in any of the manifests, including across manifests

if we detect something is in a manifest twice, or is in two different manifests, we should fail CI builds

PR #213, opened 2025-06-04
Add a castom lint rule to check for duplicates, and eslint itself.

Cross-sama:
Tried to fix issue discribed here:

https://github.com/es-tooling/module-replacements/issues/48

### DIFF
--- a/eslint-rules/no-duplicate-manifest-entries.js
+++ b/eslint-rules/no-duplicate-manifest-entries.js
@@ -1,0 +1,65 @@
+import fs from 'fs';
+import path from 'path';
+
+const manifestDir = path.resolve(process.cwd(), 'manifests');
+
+function loadManifest(fileName) {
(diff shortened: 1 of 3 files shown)


eslint-rules/no-duplicate-manifest-entries.js
```diff
@@ -0,0 +1,65 @@
...
+export default {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'disallow duplicate modules across manifests',
+    },
+    schema: [],
+  },
+  create(context) {
+    // only run once per lint run (on Program)
```
43081j, line 42: this `create` will run per JS file

we should probably run against the JSON files and report the specific file/position while traversing it

that also means we'd need to setup eslint to be able to lint JSON, though.

i.e. it should visit each manifest and check against the already seen set, to then report the exact position in the current manifest being linted

PR #225, opened 2025-08-04
feat: add string-width to preferred manifest

43081j:
Adds `string-width` since the `fast-string-width` fork is much faster.


MichaelDeBoey:
Should we add `fast-string-width` as an alternative of `unicode-string-width` as well?

PR #235, opened 2025-08-30
feat: remove `lodash.unique` & `lodash.pullby` from the manifest

outslept:
removed

- `lodash.pullby`
- `lodash.unique`

they don't exist in the registry anymore (gives 404)


43081j (approved):
they never did exist in npm. they haven't been removed, they were just never there in the first place FYI

PR #237, opened 2025-09-03
docs: add `jsx-ast-utils` alternative

hainenber:
Add `jsx-ast-utils` alternative with details.

### DIFF
--- a/docs/modules/jsx-ast-utils.md
+++ b/docs/modules/jsx-ast-utils.md
@@ -6,7 +6,7 @@
 
 ## `jsx-ast-utils-x`
 
-Practically offers similar API while having zero dependency, compared to [92 dependencies from `jsx-ast-utils`](https://npmgraph.js.org/?q=jsx-ast-utils).
+Practically offers similar API while having zero dependency and being much smaller.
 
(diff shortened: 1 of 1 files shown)


docs/modules/jsx-ast-utils.md
```diff
@@ -0,0 +1,15 @@
+# jsx-ast-utils
+
+[`jsx-ast-utils`](https://github.com/jsx-eslint/jsx-ast-utils) can be replaced with lightweight alternatives
+
+# Alternatives
+
+## `jsx-ast-utils-x`
+
+Practically offers similar API while having zero dependency, compared to [92 dependencies from `jsx-ast-utils`](https://npmgraph.js.org/?q=jsx-ast-utils).
```
43081j: i would avoid naming the exact count of dependencies since it could change over the lifetime of this doc

e.g.

> Offers a similar API with a much smaller size and zero dependencies

we probably don't really need to compare

hainenber: fixed in latest commit

43081j (approved):

PR #240, opened 2025-09-11
feat: add more entries to micro utilities manifest

### LINKED ISSUE #239: Add more micro-utilities with their replacements
| Module          | Replacement                                                                      | Weekly Downloads | NPM                                                                                                                                |
| --------------- | -------------------------------------------------------------------------------- | ---------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
| `arr-diff`      | `a.filter(x => !new Set(b).has(x))`                                              | ~15M             | [npm](https://www.npmjs.com/package/arr-diff)      |
| `array-last`    | `arr.at(-1)` (Node ≥ 16.6) or `arr[arr.length - 1]`                              | ~1.1M            | [npm](https://www.npmjs.com/package/array-last)    |
| `array-union`   | `[...new Set([...a, ...b])]`                                                     | ~47M             | [npm](https://www.npmjs.com/package/array-union)   |
| `array-unique`  | `[...new Set(arr)]`                                                              | ~14.1M           | [npm](https://www.npmjs.com/package/array-unique)  |
| `array-uniq`    | `[...new Set(arr)]`                                                              | ~8.6M            | [npm](https://www.npmjs.com/package/array-uniq)    |
| `arrify`        | `(v == null ? [] : Array.isArray(v) ? v : [v])`                                  | ~29.8M           | [npm](https://www.npmjs.com/package/arrify)        |
| `has-flag`      | `process.argv.includes('--flag')`                                                | ~204.9M          | [npm](https://www.npmjs.com/package/has-flag)      |
| `is-ci`         | `Boolean(process.env.CI)`                                                        | ~17.8M           | [npm](https://www.npmjs.com/package/is-ci)         |
| `is-finite`     | `Number.isFinite(v)`                                                             | ~5.8M            | [npm](https://www.npmjs.com/package/is-finite)     |
| `kind-of`       | `typeof v`, or `Object.prototype.toString.call(v)` to get the internal [[Class]] | ~100M            | [npm](https://www.npmjs.com/package/kind-of)       |
| `lower-case`    | `str.toLocaleLowerCase()` or `str.toLowerCase()`                                 | ~26.5M           | [npm](https://www.npmjs.com/package/lower-case)    |
| `repeat-string` | `str.repeat(n)`                                                                  | ~18.5M           | [npm](https://www.npmjs.com/package/repeat-string) |
| `slash`         | `path.startsWith('\\\\?\\') ? path : path.replace(/\\/g, '/')`                   | ~93.4M           | [npm](https://www.npmjs.com/package/slash)         |
| `uniq`          | `[...new Set(arr)]`                                                              | ~4.7M            | [npm](https://www.npmjs.com/package/uniq)          |
| `upper-case`    | `str.toLocaleUpperCase()` or `str.toUpperCase()`                                 | ~12.1M           | [npm](https://www.npmjs.com/package/upper-case)    |

outslept:
closes #239

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -3,7 +3,7 @@
     {
       "type": "simple",
       "moduleName": "arr-diff",
-      "replacement": "Use a.filter(x => !new Set(b).has(x))",
+      "replacement": "Use a.filter((item) => !b.includes(item))",
       "category": "micro-utilities"
     },
     {


manifests/micro-utilities.json
```diff
@@ -1,5 +1,41 @@
 {
   "moduleReplacements": [
+    {
+      "type": "simple",
+      "moduleName": "arr-diff",
+      "replacement": "Use a.filter(x => !new Set(b).has(x))",
```
```suggestion
      "replacement": "Use a.filter((item) => !b.includes(item))",
```

since its strict equality we can just use `includes` i think

manifests/micro-utilities.json
```diff
@@ -1,5 +1,41 @@
...
+      "category": "micro-utilities"
+    },
+    {
+      "type": "simple",
+      "moduleName": "array-last",
+      "replacement": "Use arr.at(-1) (Node ≥ 16.6) or arr[arr.length - 1]",
```
43081j: we shouldn't store version info in these messages really


```suggestion
      "replacement": "Use arr.at(-1) or arr[arr.length - 1]",
```

that is a job for docs one day

outslept: 👍🏼

manifests/micro-utilities.json
outslept: it's O(n) tho vs Set.prototype.has being O(1) average case? Do we care about such differences?

outslept: ah we need to create the set outside of the callback

```
const bSet = new Set(b);
const diff = a.filter(x => !bSet.has(x));
```

my apologies for the overlook. I guess I'll go with your suggestion

43081j: yes exactly

so i think we may as well suggest the more obvious code and people who need it to be faster can use a set

PR #252, opened 2025-09-23
feat: add `crypto-js` to the preferred manifest

### LINKED ISSUE #248: Crypto-js
# `crypto-js`

crypto-js not updated for 2 years and unmantained, so I think we should recommend people moving to native implementations of Node and Bun.

And is a library that is used quite a lot! 

Image

https://github.com/brix/crypto-js

https://github.com/brix/crypto-js?tab=readme-ov-file#discontinued

### `node:crypto` (built-in)

NodeJs already supports crypto by default with "node:crypto" so that would be my recommendation for migration

[Documentation Page](https://nodejs.org/api/crypto.html)

### `Bun.CryptoHasher` (built-in)

[Documentation Page](https://bun.com/reference/bun/CryptoHasher)

outslept:
closes #248 

there are a bunch of libs to fill the gaps, but if I list everything this would be a large encyclopedia 😄


docs/modules/crypto-js.md
```diff
@@ -0,0 +1,23 @@
...
+
+## node:crypto (built-in)
+
+Node’s official cryptography module; supports hashes/HMAC, AES-GCM, PBKDF2/scrypt, RSA/ECDSA/Ed25519, and secure RNG.
+
+[Documentation Page](https://nodejs.org/api/crypto.html)
+
+## Web Crypto API (built-in)
+
+Standards-based cryptography in browsers and modern Node (via crypto.webcrypto); async APIs for AES-GCM, HMAC, PBKDF2/HKDF, RSA/ECDSA, and secure RNG. Notable gaps: no streaming APIs, no MD5/RIPEMD160/SHA‑3, no RC4/3DES/ECB, no scrypt/Argon2 (PBKDF2/HKDF only), limited curves (P‑256/384/521; no Ed25519/Ed448 in browsers).
```
outslept: Maybe this is too much info for this. Makes it sort of hard to read

Issue #256, opened 2025-10-05
Repo: unify docs with e18e and awesome-e18e


43081j:
We currently have the following resources:

- this repo (replacement manifests, legacy docs)
- the e18e site (replacement docs)
- awesome-e18e (recommended libs/tools, including the ones in this repo)

It would be nice if we could come up with a way that all docs and replacements exist in this repo, but get synchronised into the e18e site.

similarly, if the awesome-e18e list was curated _and_ synchronised with the modules in this repo (since it needs non-replacement recommendations too - useful tools and what not).

this suggests replacements need:

- a human readable category (e.g. "CLI tools")
- a script to sync them up into the e18e site repo (probably should live in the e18e repo)

43081j:
that could work

the docs we already put in e18e are the source of truth right now, so we'd basically copy those here and overwrite the ones that exist. then update the diffs to be github-friendly, then transform them during sync like you say.

all snippets will be JSON or typescript so the `ts` language should work fine

for the slugs, you're probably right that it shouldn't be a human readable name as the consumer should decide that

gameroman:
was this solved in v3 or we want to move awesome-e18e here too?

43081j:
the awesome list is still poorly maintained. i do agree now that we should just automate it if we can or archive it

PR #258, opened 2025-10-23
docs: transported docs from e18e; slight stylistic cleanup

outslept:
related #256


docs/modules/eslint-plugin-react.md
```diff
@@ -1,40 +1,36 @@
...
+# Replacements for `eslint-plugin-react`
 
 ## `@eslint-react/eslint-plugin`
 
-Not a drop-in replacement but a feature rich alternative with many of the same rules.
-
-[Project Page](https://github.com/Rel1cx/eslint-react)
-[npm](https://www.npmjs.com/package/@eslint-react/eslint-plugin)
-
-## `@stylistic/eslint-plugin-jsx`
```
43081j, line 14: we've lost the stylistic recommendation

we don't need to exhaustively list the rules like it is here. we just need to explain it should be used for stylistic rules the react plugin was previously used for

docs/modules/find-up.md
```diff
@@ -1,25 +1,74 @@
...
+```ts
+import * as find from 'empathic/find' // [!code ++]
+import { findUp } from 'find-up' // [!code --]
 
-[npm](https://www.npmjs.com/package/empathic)
+await findUp('package.json') // [!code --]
+find.up('package.json') // [!code ++]
+```
 
-## pkg-types
```
43081j, line 17: we've lost this one too

pkg-types is a good option if you want to parse and strongly type the package.json in one go

docs/modules/globby.md
```diff
@@ -0,0 +1,42 @@
...
+
+# Replacements for `globby`
+
+## `tinyglobby`
+
+[`globby`](https://github.com/sindresorhus/globby) is a convenience wrapper around [`fast-glob`](https://github.com/mrmlnc/fast-glob).
+
+You can follow roughly the same migration process as documented in the [`fast-glob`](./fast-glob.md) replacement guide, since `globby` is built on top of it and the main differences are its extra conveniences.
+
+If you don’t need `.gitignore` handling, prefer [`tinyglobby`](https://github.com/SuperchupuDev/tinyglobby/) - it’s smaller and faster. If you do need `.gitignore` behavior, pair `tinyglobby` with a small git-based helper. For most cases, this will likely be good enough.p
```
43081j:
```suggestion
If you don’t need `.gitignore` handling, prefer [`tinyglobby`](https://github.com/SuperchupuDev/tinyglobby/) - it’s smaller and faster. If you do need `.gitignore` behavior, pair `tinyglobby` with a small git-based helper. For most cases, this will likely be good enough:
```

docs/modules/momentjs.md
```diff
@@ -1,36 +0,0 @@
...
-[npm](https://www.npmjs.com/package/dayjs)
-
-## date-fns
-
-Date-fns offers a series of functions for manipulating JavaScript [`Date`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Date) objects. For more details, scroll to "Why date-fns?" on the date-fns home page.
-
-[Project Page](https://github.com/date-fns/date-fns)
-[npm](https://www.npmjs.com/package/date-fns)
-
-## You don’t (may not) need Moment.js
```
43081j, line 32: we lost this one

docs/modules/eslint-plugin-react.md
outslept: I think this was intentional as migration guide covers where you need eslint-plugin-react-x and where you need stylistic. Don't think we need to duplicate the info

docs/modules/globby.md
outslept: 👍🏼

docs/modules/momentjs.md
outslept: I'm a bit skeptical of this. This would be the 4th anchor and the page itself is just recipes for migrating to date-fns/luxon/day.js/Temporal API. Ideally you won't use any of the mentioned libraries and stick to Date or Temporal (when it's stable enough)

docs/modules/find-up.md
outslept: 👍🏼

docs/modules/find-up.md
```diff
@@ -1,25 +1,88 @@
...
+)
+```
+
+## `pkg-types`
+
+[`pkg-types`](https://github.com/unjs/pkg-types) provides utilities for reading and writing package.json, tsconfig.json, and other configuration files with TypeScript support.
+
+```ts
+import { findUp } from 'find-up' // [!code --]
+import { readPackage } from 'read-pkg' // [!code --]
```
43081j: i think we only need to show one, as real code wouldn't be importing both of these. its a realistic diff remember

outslept: 👍🏼

docs/modules/tempy.md
```diff
@@ -1,42 +1,44 @@
...
+## Deno
+
+Deno provides built-in [`Deno.makeTempDir`](https://docs.deno.com/api/deno/~/Deno.makeTempDir) and [`Deno.makeTempFile`](https://docs.deno.com/api/deno/~/Deno.makeTempFile) for creating unique temporary directories and files in the system temp directory (or a custom `dir`). You can also set `prefix` and `suffix`. Both return the full path and require `--allow-write`.
+
+```ts
+import { temporaryDirectory } from 'tempy' // [!code --]
+
+const tempDir = temporaryDirectory({ prefix: 'foo-' }) // [!code --]
+
+const tempDir = await Deno.makeTempDir({ prefix: 'foo-' }) // [!code ++]
```
43081j, line 31:
```suggestion
import { temporaryDirectory } from 'tempy' // [!code --]

const tempDir = temporaryDirectory({ prefix: 'foo-' }) // [!code --]
const tempDir = await Deno.makeTempDir({ prefix: 'foo-' }) // [!code ++]
```

docs/modules/tempy.md
```diff
@@ -1,42 +1,44 @@
...
+```
+
+```ts
+import { temporaryFile } from 'tempy' // [!code --]
+
+const tempFile = temporaryFile({ extension: 'txt' }) // [!code --]
+
+const tempFile = await Deno.makeTempFile({ suffix: '.txt' }) // [!code ++]
```
43081j, line 38:
```suggestion
import { temporaryFile } from 'tempy' // [!code --]

const tempFile = temporaryFile({ extension: 'txt' }) // [!code --]
const tempFile = await Deno.makeTempFile({ suffix: '.txt' }) // [!code ++]
```

docs/modules/eslint-plugin-react.md
43081j: which migration guide?

this one here doesn't mention stylistic unless im being blind

outslept: we have a link in the blockquote, which leads us to a migration guide. and all this is covered there

https://github.com/outslept/module-replacements/blob/9a8e364a12da879d8cabd67303ccce38d3ee9fe9/docs/modules/eslint-plugin-react.md?plain=1#L36

PR #266, opened 2025-12-10
feat: schema v3 iteration

43081j:
This is a first attempt at the new version of the schema.

The major change here is that we'd go from having an array of many
replacements, to one entry per source module with many replacements.

For example:

```ts
{
  "type": "module",
  "moduleName": "chalk",
  "replacements": [
    {
      "type": "native",
      "engines": [
        { "engine": "node", "minVersion": "X.Y.Z" }
      ],
      "url": { "type": "node", "id": "styleText" },
      "description": "The `styleText` built-in provides [...]"
    }
  ]
}
```

### DIFF
--- a/manifests/native.json
+++ b/manifests/native.json
@@ -480,6 +480,18 @@
         "compatKey": "javascript.builtins.Date"
       }
     },
+    "Error.prototype.cause": {
+      "id": "Error.prototype.cause",
+      "type": "native",
+      "url": {
+        "type": "mdn",
+        "id": "Web/JavaScript/Reference/Global_Objects/Error/cause"
+      },
+      "webFeatureId": {
+        "featureId": "error-cause",
+        "compatKey": "javascript.builtins.Error.cause"
+      }
+    },
     "Function.prototype.bind": {
       "id": "Function.prototype.bind",
       "type": "native",
@@ -925,6 +937,18 @@
         "compatKey": "javascript.builtins.Reflect.ownKeys"
       }
     },
+    "RegExp.prototype.flags": {
+      "id": "RegExp.prototype.flags",
+      "type": "native",
+      "url": {
+        "type": "mdn",
+        "id": "Web/JavaScript/Reference/Global_Objects/RegExp/flags"
+      },
+      "webFeatureId": {
+        "featureId": "regexp",
+        "compatKey": "javascript.builtins.RegExp.flags"
+      }
+    },
     "String.prototype.at": {
       "id": "String.prototype.at",
       "type": "native",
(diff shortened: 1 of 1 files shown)


43081j:
i now added sync scripts to use runtime compat data, and web feature data

so we have this `engines` populated from the web features data:

```ts
        {
          "type": "native",
          "webFeatureId": "fetch",
          "engines": [
            {
              "engine": "chrome",
              "minVersion": "42"
            },
            {
              "engine": "chrome_android",
              "minVersion": "42"
            },
            {
              "engine": "edge",
              "minVersion": "14"
            },
            {
              "engine": "firefox",
              "minVersion": "39"
            },
            {
              "engine": "firefox_android",
              "minVersion": "39"
            },
            {
              "engine": "safari",
              "minVersion": "10.1"
            },
            {
              "engine": "safari_ios",
              "minVersion": "10.3"
            }
          ],
          "url": {
            "type": "mdn",
            "id": "Web/API/Fetch_API"
          },
          "description": "Use native fetch"
        },
```

and this from the runtime data:

```ts
        {
          "type": "native",
          "nodeFeatureId": {
            "moduleName": "node:buffer",
            "exportName": "Buffer"
          },
          "engines": [
            {
              "engine": "cloudflare",
              "minVersion": "24.9.0"
            },
            {
              "engine": "deno",
              "minVersion": "24.9.0"
            },
            {
              "engine": "vercel",
              "minVersion": "24.9.0"
            },
            {
              "engine": "netlify",
              "minVersion": "24.9.0"
            }
          ],
          "url": {
            "type": "e18e",
            "id": "buf-compare"
          },
          "description": "Use native Buffer.compare"
        }
```

43081j:
so i dont forget:

- we should have two top-level collections
  - one is the `replacements` (mappings of module to its replacements)
  - one is the targets (things we replace those modules with, i.e. what we mapped to)
- a replacement mapping's modules are just a set of IDs we later lookup from the targets list
- we need some representation of which replacements are "preferred"

scripts/update-engines-runtimes.ts
```diff
@@ -0,0 +1,222 @@
+import {readdir, readFile, writeFile} from 'node:fs/promises';
+import {fileURLToPath} from 'node:url';
+import * as path from 'node:path';
+import type {
+  ManifestModule,
+  ModuleReplacementDescriptor,
+  EngineConstraint
+} from '../src/types.js';
+
+const scriptDir = fileURLToPath(new URL('.', import.meta.url));
```
ghostdevv, line 11: import.meta.dirname 😔

src/types.ts
```diff
@@ -1,36 +1,63 @@
...
+  replacementModule: string;
 }
 
 export interface NativeModuleReplacement extends ModuleReplacementLike {
   type: 'native';
-  mdnPath: string;
-  nodeVersion: string;
-  replacement: string;
+  url: KnownUrl;
+  description: string;
```
ghostdevv: I think I already said this somewhere in Discord, but currently this description is the syntax to replace with - so for example the replacement `Number.isNaN` is

```json
{
	"id": "Number.isNaN",
	"type": "native",
	"description": "Number.isNaN",
  ...
}
```

It'd be handy if the actual syntax replacement and description was separate, for example:

```json
{
	"id": "Number.isNaN",
	"type": "native",
	"description": "Safer as doesn't attempt to coerce input to a number",
	"with": [
		"Number.isNaN"
	],
  ...
}
```

Could even take this further and provide ast-grep

```json
{
  ...,
	"with": [
		{ "option": "Number.isNaN", "match": "isNaN($ARG)", "replace": "Number.isNaN($ARG)" }
	]
}
```

although that won't account for the whole story cases where we're replacing the `number.isnan` package for example

43081j: matching rules will get pretty big so im not sure its a good idea to keep them in the manifests. thats what the `module-replacements-codemods` repo is for, for example.

the plan here was just to have a new `example` property that contains an example of using the replacement (optional since some would be too big examples)

the description is wrong for a bunch of these right now and needs expanding like you said

PR #271, opened 2025-12-26
Add portal-vue to module replacements

### LINKED ISSUE #270: portal-vue
# `portal-vue`

Portal-vue is a package that was created to add portals to vue ( useful for dialogs,modals etc ) this package was created when vue 2 didn't support portals (which in vue are called teleport)

The last update was 3 years ago and even if they also support vue 3 the best option is to just migrate to the native implementation which is teleport.

https://www.npmjs.com/package/portal-vue

---
description: vue 3 already supports teleport natively (portal-vue) making the extra dependency unnecessary 
---

App.vue
```
<template>
  <div>
    <h1>Nuxt App</h1>

    <portal to="teleport-target">
      <p>This is portal</p>
    </portal>
  </div>
</template>
```

```
<div id="portal-target">
  <!-- Portal content will render here -->
</div>
```

# Replacements for `portal-vue`

## vue native `teleport` API

Example: 

App.vue
```
<template>
  <div>
    <h1>Nuxt App</h1>

    <teleport to="#teleport-target">
      <p>This is teleported</p>
    </teleport>
  </div>
</template>
```

Somewhere in the app
```
    <div id="teleport-target">
      <!-- Teleported content goes here -->
    </div>

```

joaopedrodcf:
First contribution to the project :D

Closes #270

### DIFF
--- a/manifests/preferred.json
+++ b/manifests/preferred.json
@@ -2270,12 +2270,6 @@
     },
     {
       "type": "documented",
-      "moduleName": "portal-vue",
-      "docPath": "portal-vue",
-      "category": "preferred"
-    },
-    {
-      "type": "documented",
       "moduleName": "pkg-dir",
       "docPath": "pkg-dir",
       "category": "preferred"
@@ -2284,6 +2278,12 @@
       "type": "documented",
       "moduleName": "pkg-up",
       "docPath": "find-up",
+      "category": "preferred"
+    },
+    {
+      "type": "documented",
+      "moduleName": "portal-vue",
+      "docPath": "portal-vue",
       "category": "preferred"
     },
     {


docs/modules/portal-vue
```diff
@@ -0,0 +1,27 @@
+---
+description: Portal-vue was used to Teleport a part of component's template into a DOM node that exists outside of the hierarchy and Vue 3 supports this feature with Teleport api (There are still situations )
```
43081j: this description is used more like a sub-title in embeds, etc (imagine a bsky embed).

so we should put something more concise here that's less about "why" and more about "what" (see the other files for examples)

joaopedrodcf: Yeah my bad with the descritiption, should have check the rest for cosistency, its updated.

what do u think ?

43081j: looks good to me 👍

PR #272, opened 2025-12-26
Add node-fetch and cross-fetch

### LINKED ISSUE #269: node-fetch
# `node-fetch`

node-fetch was a package created for fetch to run in node ( similar to cross-fetch but cross-fetch is to work equal in browser and server ), it has still millions of downloads even if fetch since 18 ( experimental ) and 21 ( stable ) supports fetch API

Image

https://www.npmjs.com/package/node-fetch

---
description: Node Js supports fetch API since v21 and experimental with a flag since 18
---

# Replacements for `node-fetch`

## Native `fetch` API

The native [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) API is available in Node.js (since v18.x) and all modern browsers. For most HTTP requests, it can replace `axios` without extra dependencies.

Example:

```ts
// GET
const res = await fetch('https://api.example.com/data')
const data = await res.json()

// POST
await fetch('https://api.example.com/data', {
  method: 'POST',
  headers: { 'Content-Type': 'application/json' },
  body: JSON.stringify({ key: 'value' })
})
```

## `ky`

[`ky`](https://github.com/sindresorhus/ky) is a lightweight HTTP client based on the Fetch API with timeout support, hooks (interceptors) and other helpers.

Example:

```ts
import ky from 'ky'

const api = ky.create({
  prefixUrl: 'https://api.example.com',
  timeout: 5000, // ms
})

const data = await api.get('users').json()
```

## `ofetch`

[`ofetch`](https://github.com/unjs/ofetch) s a fetch wrapper with automatic JSON parsing, request/response interceptors, and retries.

Example:

```ts
import { ofetch } from 'ofetch'

const api = ofetch.create({
  baseURL: 'https://api.example.com',
})

const data = await api('/user', { query: { id: 123 } })
const created = await api('/items', { method: 'POST', body: { name: 'A' } })
```

joaopedrodcf:
Adding node-fetch and cross-fetch to the module replacements, with this we have modules that will require the same replacements how should we share that part ? Should we do it like lodash/underscore example ?

I also created issues in the respective repositories as it would be nice to warn users about not needing the modules depending on their engines:
https://github.com/node-fetch/node-fetch/issues/1857
https://github.com/lquixada/cross-fetch/issues/209

Closes #269  #268


docs/modules/node-fetch.md
```diff
@@ -0,0 +1,58 @@
+---
+description: node-fetch was a package created for fetch to run in node (similar to cross-fetch but cross-fetch is to work equal in browser and server), it has still millions of downloads even if fetch since 18 (experimental) and 21 (stable) supports fetch API
```
43081j: the description is used for embeds etc, its meant to be more concise. see the axios one for example:

```md
---
description: Modern alternatives to the axios package for making HTTP requests in browsers and Node.js
---
```

43081j:
i wonder if we should just have a `fetch.md` which axios, node-fetch and cross-fetch link to.

we generally do one-to-one docs so we can show migration examples (rich diffs). but in this case, there are no diffs, rather just basic examples. so we probably could have one doc

thoughts? would be nice to get the opinion of someone like @outslept or @Fuzzyma too

joaopedrodcf:
Okay will do it tomorrow guys 💪

joaopedrodcf:
Sorry for the delay in such a small change is done @43081j  and @Fuzzyma

joaopedrodcf:
@43081j @Fuzzyma fixed the sorting that was causing the CI to fail, didn't noticed it the other day! :D

PR #273, opened 2026-01-05
feat: bundle manifests in code instead of loading from fs

ghostdevv:
I realised when deploying code that imported `module-replacements` to Cloudflare that it was reading the manifests from the file system, which of course didn't work. This is the minimum change I could get to instead have the json become javascript without_ removing tshy or introducing a breaking change. I couldn't find if this had been talked about before, so apologies for the impromptu PR


tsconfig.json
```diff
@@ -9,7 +9,9 @@
     "forceConsistentCasingInFileNames": true,
     "strict": true,
     "noUnusedLocals": true,
-    "noUnusedParameters": true
+    "noUnusedParameters": true,
+    "resolveJsonModule": true,
```
43081j: we probably don't need this now that no json gets imported, right?

scripts/manifests-to-ts.js
```diff
@@ -0,0 +1,30 @@
...
+for (const manifest of manifests) {
+  if (manifest.isFile() && path.extname(manifest.name) === '.json') {
+    const content = await readFile(
+      path.join(jsonManifestsPath, manifest.name),
+      'utf8'
+    );
+
+    await writeFile(
+      path.join(tsManifestsPath, manifest.name.replace(/\.json$/, '.ts')),
+      `import {ManifestModule} from '../types.js';\n\nexport default ${content.trim()} as ManifestModule;\n`
```
43081j: it'd be better to type it rather than cast it

basically:

```ts
const manifest: ManifestModule = ${content.trim()};
export default manifest;
```

tsconfig.json
ghostdevv: ah, that's an artifact from my first seeing if tshy would bundle json imports

scripts/manifests-to-ts.js
ghostdevv: ah yes!! also an artifact from my testing whether tshy would bundle json - in that case the cast was necessary because typescript widened the `type` field :melting_face:

ghostdevv:
> i think this makes sense but it probably needs to happen in the next major since we have the json files in our package `exports`

They're still there - from a consumer point this shouldn't be a breaking change as only how the `main.ts` imports the manifests has changed. Before, read from disk at runtime, now _simply_ imported.

> maybe you can help me get the new manifest structure over the line in #266 👀

:eyes:

gameroman:
I think he means to remove the `.json` export from `package.json`

bluwy:
> how come?

I mean this PR, which makes the data exist in TS and JSON. 

> rewrite the manifests as typescript, yes or no?

I don't really mind either as long as it's only in TS or JSON, and it's possible to only import a single type of replacement data (not load the others in memory).

I have a slight preference to JSON though if the data would always be serializable, and would be easier to be consumed by non-JS/native tooling.

bluwy:
I think we can use import attributes but might want to set an explicit node engines range for that support. I believe `cli-boxes` have a similar setup.

43081j:
it turns out cloudflare can't do import attributes. so we would have to decide if that's ok or not.

most workers will be bundled so its probably not a problem. but just FYI

PR #274, opened 2026-01-22
feat: add cli arg parser replacements

bjnewman:
## Summary

Adds migration documentation for replacing CLI argument parsing packages with Node.js built-in `util.parseArgs` (available in Node 18.3+/16.17+).

### Packages covered
- `minimist`
- `mri`
- `arg`
- `meow`
- `yargs-parser`
- `yargs`
- `commander`
- `sade`

### Changes
- New `docs/modules/parseargs.md` with migration examples for each package
- Added 8 entries to `manifests/preferred.json`

### Notes
- Examples use nullish coalescing (`??`) for defaults instead of `parseArgs`'s `default` option to ensure compatibility with all Node 18.3+ versions
- Includes feature comparison table and caveats about subcommands/auto-help
- Covers advanced patterns like token-based parsing for minimist compatibility


ljharb:
Also worth suggesting as an alternative is https://npmgraph.js.org/?q=pargs, which is built on top of util.parseArgs, has no deps, is written in ESM, and adds a number of helpful features when migrating from the listed packages.

43081j:
sorry its taken me a while to get to this

i think we need some 2nd opinions here as this diverges from the usual structure we have

we follow one of two paths so far:

- the document is named after the thing being replaced, and lists migrations to the various options
- the document is named after the primary replacement (e.g. `fetch`), and lists other options _but usually no migrations_

in this case, it is listing the things being replaced along with their migrations. which is different to all other docs

@Fuzzyma thoughts?

Fuzzyma:
hm, so would you rather have x documents that all have one migration "x to fetch" and then in the fetch document link to all migrations? I think that would be cumbersome.

Is there a reason we usually dont have migrations in the "fetch-like" documents?

bjnewman:
Thanks for the great work on this project and for taking the time to
review! I went with a single doc covering all 8 packages because their
APIs all converge on the same built-in replacement (`util.parseArgs`),
and I wanted to show the specific before/after for each since they have
pretty different APIs.

I was loosely modeling off of `fetch.md`, `lodash-underscore.md`, and
`bluebird-q.md`, which also group multiple packages into one doc. But I
can see how the per-package migration examples diverge from those.

Happy to restructure to better fit the project — just let me know what
you'd prefer and I'll update.

gameroman:
Yeah, I think we should have separate file for each of these just how we have with `glob` etc

manifests/preferred.json
```diff
@@ -3259,6 +3307,13 @@
       "nodeFeatureId": {"moduleName": "node:util", "exportName": "types"},
       "url": {"type": "node", "id": "api/util.html#utiltypes"}
     },
+    "util.parseArgs": {
+      "id": "util.parseArgs",
+      "type": "native",
+      "description": "util.parseArgs",
```
gameroman: Can be removed

```suggestion
```

gameroman (changes requested):
Added one comment. Everything else looks fine

gameroman (approved):

43081j (dismissed):

manifests/preferred.json
```diff
@@ -2464,6 +2494,12 @@
       "replacements": ["fs.rm", "fs.rmdir", "premove"],
       "url": {"type": "e18e", "id": "rimraf"}
     },
+    "sade": {
```
43081j: I'm not sure sade should be in here. 

From what I remember it uses mri under the hood to do arguments parsing. But it exists to provide help text etc on top of that. 

In a future pr, we should probably add sade as a replacement itself for some packages in fact

gameroman: It does use `mri`, yes

joaopedrodcf (approved):

Dustin4444:
> Added one comment. Everything else looks fine

manifests/preferred.json
43081j: that's why it shouldn't really be here.

sade exists to provide a CLI args parser _and_ help text generator (the former being provided by mri under the hood).

parseArgs replaces mri, but doesn't replace sade.

ghostdevv: did we decide whether we're removing this here, and adding it in the follow up?

ghostdevv (changes requested):

43081j (dismissed):

43081j:
@ghostdevv i've reworked this to only have the option parsers, and added `mri` as a replacement choice too (for anyone who doesn't have new enough Node for `parseArgs`)

gameroman (approved):

ghostdevv (dismissed):
LGTM - I updated the branch and tweaked the title @43081j

43081j (approved):

PR #283, opened 2026-02-09
chore: update gzip-size typo

bluwy:
I think it's missing a `.length`


docs/modules/gzip-size.md
```diff
@@ -45,6 +45,6 @@ import { gzipSizeFromFileSync } from 'gzip-size'; // [!code --]
 
 const path = '/path/to/file';
 
-console.log(gzipSync(readFileSync(path))); // [!code ++]
+console.log(gzipSync(readFileSync(path)).length); // [!code ++]
```
43081j: while you're here can you also switcheroo them

`--` lines should come first, `++` lines after

PR #284, opened 2026-02-09
style: format more files

bluwy:
Format more files, especially the `docs/modules/*.md` files. While I was editing the markdown, I feel it would be nice if formatting is enabled for it.

I added a prettier config override for these markdown file to reduce the diff.


manifest-schema.json
```diff
@@ -23,11 +23,7 @@
                 "type": "string"
```
43081j, line 23: iirc we generate this file so it may not make sense formatting it

it is committed to the repo but is generated from the typescript types from what i remember

43081j (approved):

PR #294, opened 2026-02-20
feat: descriptions and examples for simple replacements

paoloricciuti:
This adds better descriptions and examples to the `simple` replacements

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -40,7 +40,7 @@
       "id": "snippet::array-last",
       "type": "simple",
       "description": "You can use arr.at(-1) if supported or arr[arr.length - 1] to get the last element of an array.",
-      "example": "const last = (arr) => arr.at(-1) ?? arr[arr.length - 1]"
+      "example": "const last = (arr) => arr.at(-1);\n// or in older environments\nconst last = (arr) => arr[arr.length - 1]"
     },
     "snippet::array-union": {
       "id": "snippet::array-union",


manifests/micro-utilities.json
```diff
@@ -27,172 +27,206 @@
...
-      "description": "Use a.filter((item) => !b.includes(item))"
+      "description": "You can use a combination of filter and includes to calculate the difference between two arrays.",
+      "example": "const difference = (a, b) => a.filter((item) => !b.includes(item))"
     },
     "snippet::array-last": {
       "id": "snippet::array-last",
       "type": "simple",
-      "description": "Use arr.at(-1) or arr[arr.length - 1]"
+      "description": "You can use arr.at(-1) if supported or arr[arr.length - 1] to get the last element of an array.",
+      "example": "const last = (arr) => arr.at(-1) ?? arr[arr.length - 1]"
```
```suggestion
      "example": "const last = arr.at(-1)"
```

i would just do this since you'd never really combine both. its just to show a usage example

manifests/micro-utilities.json
```diff
@@ -27,172 +27,206 @@
...
       "type": "simple",
-      "description": "Use [...new Set([...a, ...b])]"
+      "description": "You can use a combination of the spread operator and Set to create a union of two arrays.",
+      "example": "const union = (a, b) => [...new Set([...a, ...b])]"
     },
     "snippet::array-unique": {
       "id": "snippet::array-unique",
       "type": "simple",
-      "description": "Use [...new Set(arr)]"
+      "description": "You can create a new Set from your array to remove duplicates and then spread it back into an array.",
```
```suggestion
      "description": "You can convert to and from a Set to remove duplicates from an array.",
```

manifests/micro-utilities.json
```diff
@@ -27,172 +27,206 @@
...
       "type": "simple",
-      "description": "Use Object.prototype.toString.call(v) === \"[object Boolean]\""
+      "description": "You can use typeof to check if a value is a boolean primitive, and Object.prototype.toString.call to check if it's a Boolean object.",
+      "example": "Object.prototype.toString.call(v) === \"[object Boolean]\""
     },
     "snippet::is-ci": {
       "id": "snippet::is-ci",
       "type": "simple",
-      "description": "Use Boolean(process.env.CI)"
+      "description": "Every major CI provider sets a unique environment variable that you can check for to detect if you're running in a CI environment.",
```
```suggestion
      "description": "Every major CI provider sets a CI environment variable that you can use to detect if you're running in a CI environment.",
```

manifests/micro-utilities.json
```diff
@@ -27,172 +27,206 @@
...
       "type": "simple",
-      "description": "Use v instanceof Date, or if cross-realm, use Object.prototype.toString.call(v) === \"[object Date]\""
+      "description": "You can use `instanceof Date`, or if cross-realm, use Object.prototype.toString.call(v) === \"[object Date]\"",
+      "example": "const isDate = v instanceof Date;\n// for cross-realm\nconst isDateCrossRealm = Object.prototype.toString.call(v) === \"[object Date]\""
     },
     "snippet::is-even": {
       "id": "snippet::is-even",
       "type": "simple",
-      "description": "Use (n % 2) === 0"
+      "description": "You can use the module operator to check if a number is even.",
```
```suggestion
      "description": "You can use the modulo operator to check if a number is even.",
```

manifests/micro-utilities.json
```diff
@@ -27,172 +27,206 @@
...
       "type": "simple",
-      "description": "Use v && typeof v === \"object\" && (Object.getPrototypeOf(v) === null || Object.getPrototypeOf(v) === Object.prototype)"
+      "description": "You can use `typeof` to check if a value is an object and `Object.getPrototypeOf` to ensure it's a plain object.",
+      "example": "const isObject = (obj) => obj && typeof obj === \"object\" && (Object.getPrototypeOf(obj) === null || Object.getPrototypeOf(obj) === Object.prototype);"
     },
     "snippet::is-odd": {
       "id": "snippet::is-odd",
       "type": "simple",
-      "description": "Use (n % 2) === 1"
+      "description": "You can use the module operator to check if a number is odd.",
```
```suggestion
      "description": "You can use the modulo operator to check if a number is odd.",
```

manifests/micro-utilities.json
```diff
@@ -27,172 +27,206 @@
...
-      "description": "Use (\"TRAVIS\" in process.env)"
+      "description": "You can check if the current environment is Travis CI by checking if the `TRAVIS` environment variable is set.",
+      "example": "const isTravis = () => \"TRAVIS\" in process.env;"
     },
     "snippet::is-whitespace": {
       "id": "snippet::is-whitespace",
       "type": "simple",
-      "description": "Use str.trim() === \"\" or /^\\s*$/.test(str)"
+      "description": "You can check if a string contains only whitespace with RegExp or by trimming it and comparing it to an empty string.",
+      "example": "const isWhitespace = (str) => str.trim() === \"\" || /^\\s*$/.test(str);"
```
```suggestion
      "example": "const isWhitespace = (str) => /^\\s*$/.test(str);"
```

similar to the other i commented on, we should just show a real usage. so in this case, we probably prefer the regex to since it'll be faster

manifests/micro-utilities.json
paoloricciuti: Yeah this was actually AI autocorrecting and slipping in (I fucking hate autocorrection...so invasive). I guess we could add a comment like in the `cross-realm` ones just to show both examples?

manifests/micro-utilities.json
paoloricciuti: Hahahaha that's a funny typo

manifests/micro-utilities.json
paoloricciuti: Oh fuck now it seems like I didn't know it was called `modulo` 😅

manifests/micro-utilities.json
paoloricciuti:
```suggestion
      "example": "const last = (arr) => arr.at(-1);\n// or in older environments\nconst last = (arr) => arr[arr.length - 1]"
```

Something like this...wdyt?

43081j: that looks good to me 👍

PR #295, opened 2026-02-20
feat: add descriptions to `removal` type

paoloricciuti:
Add the descriptions for the `removal` replacements.

### DIFF
--- a/manifests/native.json
+++ b/manifests/native.json
@@ -1108,7 +1108,7 @@
     "es-errors": {
       "id": "es-errors",
       "type": "removal",
-      "description": "Error, EvalError, RangeError, ReferenceError, SyntaxError, TypeError or URIError that this package shim are generally available in modern environments.",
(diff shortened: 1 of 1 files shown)


manifests/native.json
```diff
@@ -1108,7 +1108,7 @@
     "es-errors": {
       "id": "es-errors",
       "type": "removal",
-      "description": "TODO",
+      "description": "Error, EvalError, RangeError, ReferenceError, SyntaxError, TypeError or URIError that this package shim are generally available in modern environments.",
```
43081j: i wouldn't bother exhaustively listing them. i'd just say something like "Error and its subclasses are natively built into all modern environments"

paoloricciuti:
```suggestion
      "description": "Error and its subclasses are natively built into all modern environments.",
```

43081j (approved):

PR #311, opened 2026-02-22
add `is-buffer` to replacements

### LINKED ISSUE #307: `is-buffer`
## Before

```js
const isBuffer = require('is-buffer')
isBuffer(obj)
```

## After

```js
Buffer.isBuffer(obj)
```

gameroman:
Closes #307

### DIFF
--- a/manifests/native.json
+++ b/manifests/native.json
@@ -447,7 +447,6 @@
     "Buffer.from": {
       "id": "Buffer.from",
       "type": "native",
-      "description": "Use global `Buffer.from` in Node.js",
       "url": {
         "type": "node",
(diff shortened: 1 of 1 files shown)


manifests/native.json
```diff
@@ -456,6 +456,18 @@
         "moduleName": "buffer"
       }
     },
+    "Buffer.isBuffer": {
+      "id": "Buffer.isBuffer",
+      "type": "native",
+      "description": "Use global `Buffer.isBuffer` in Node.js",
```
43081j: we don't actually need a description since these will just link to the docs

we probably should remove the one from `Buffer.from` that got added too

PR #312, opened 2026-02-22
add `safe-buffer` to replacements

### LINKED ISSUE #305: `safe-buffer`
## Before

```js
const Buffer = require('safe-buffer').Buffer;
const buf = new Buffer('hello world', 'utf8');
```

## After

```js
const buf = new Buffer('hello world', 'utf8');
```

gameroman:
Closes #305

### DIFF
--- a/manifests/native.json
+++ b/manifests/native.json
@@ -1220,7 +1220,7 @@
     "safe-buffer": {
       "id": "safe-buffer",
       "type": "removal",
-      "description": "You don't need this dependency anymore.",
+      "description": "All modern versions of Node have `Buffer.from` for safely creating buffers from arbitrary types.",
       "url": {"type": "node", "id": "api/buffer.html"}
     }
   },


manifests/native.json
```diff
@@ -1216,6 +1216,12 @@
         "featureId": "number",
         "compatKey": "javascript.builtins.parseInt"
       }
+    },
+    "safe-buffer": {
+      "id": "safe-buffer",
+      "type": "removal",
+      "description": "You don't need this dependency anymore.",
```
43081j:
```suggestion
      "description": "All modern versions of Node have Buffer APIs natively.",
```

gameroman: It's more about it being safe now, not about being available

43081j: ok i see. its because today you can use `Buffer.from` and not need this package anymore

maybe:

```suggestion
      "description": "All modern versions of Node have `Buffer.from` for safely creating buffers from arbitrary types.",
```

PR #315, opened 2026-02-23
add `base64-js` to replacements

### LINKED ISSUE #302: `base64-js`
### Before

```js
import { fromByteArray } from 'base64-js'
const base64 = fromByteArray(bytes)
```

### After

```js
const base64 = Buffer.from(bytes).toString('base64')

// or

const base64 = bytes.toBase64()

// or

const base64 = btoa(String.fromCodePoint(...bytes))
```

### Before

```js
import { toByteArray } from 'base64-js'
const bytes = toByteArray(base64)
```

### After

```js
const bytes = Buffer.from(base64, 'base64')

// or

const bytes = Uint8Array.from(Buffer.from(base64, 'base64'))

// or

const bytes = Uint8Array.fromBase64(base64)

// or

const bytes = Uint8Array.from(atob(base64), c => c.charCodeAt(0))
```

### Before

```js
import { byteLength } from 'base64-js'
byteLength(base64)
````

### After

```js
Buffer.from(base64, 'base64').byteLength

// or

Uint8Array.fromBase64(base64).byteLength
```

gameroman:
Closes #302

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -54,8 +54,8 @@
       "description": "You can convert to and from a Set to remove duplicates from an array.",
       "example": "const unique = (arr) => [...new Set(arr)]"
     },
-    "snippet::base64-js": {
-      "id": "snippet::base64-js",
+    "snippet::base64": {
+      "id": "snippet::base64",
       "type": "simple",
       "description": "Every modern runtime provides a way to convert byte array to and from base64.",
       "example": "// From base64 to Uint8Array\nconst bytes = Uint8Array.fromBase64(base64)\n// From Uint8Array to base64\nconst base64 = bytes.toBase64()"
@@ -275,7 +275,7 @@
     "base64-js": {
       "type": "module",
       "moduleName": "base64-js",
-      "replacements": ["snippet::base64-js"]
+      "replacements": ["snippet::base64"]
     },
     "call-bind": {
       "type": "module",


manifests/micro-utilities.json
```diff
@@ -54,6 +54,12 @@
       "description": "You can convert to and from a Set to remove duplicates from an array.",
       "example": "const unique = (arr) => [...new Set(arr)]"
     },
+    "snippet::base64-js": {
```
43081j: it isn't base64-js specific. you're showing that for all base64 tasks, regardless of library, here's how you can encode

manifests/micro-utilities.json
```diff
@@ -54,6 +54,12 @@
...
+      "id": "snippet::base64-js",
+      "type": "simple",
+      "description": "Every modern runtime provides a way to convert byte array to and from base64.",
+      "example": "// fromByteArray\nbytes.toBase64();\n// toByteArray\nUint8Array.fromBase64(base64);"
```
43081j: similarly, here we shouldn't mention `fromByteArray` / `toByteArray`.

this is the generic entry for "how to convert to base64", it isn't specific to base64-js.

better to say "Convert to base64" and "Convert from base64"

but what is `bytes`? its unclear

manifests/micro-utilities.json
gameroman: I think we can probably keep this name for now, and change it later if we add another package that has similar replacement?

manifests/micro-utilities.json
gameroman: Updated

manifests/micro-utilities.json
43081j: i would just call it `base64` or something generic even now.

we want to be consistent with the other replacements, in that none are named after what they replace

gameroman: Updated

PR #317, opened 2026-02-23
add `array-initial` to replacements

### LINKED ISSUE #296: `array-initial`
## Before

```js
const initial = require('array-initial')

initial(array, n)
```

## After

```js
array.slice(0, array.length - n)
```

gameroman:
Closes #296

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -36,17 +36,17 @@
       "description": "You can use a combination of filter and includes to calculate the difference between two arrays.",
       "example": "const difference = (a, b) => a.filter((item) => !b.includes(item))"
     },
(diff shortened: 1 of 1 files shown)


manifests/micro-utilities.json
```diff
@@ -36,6 +36,12 @@
       "description": "You can use a combination of filter and includes to calculate the difference between two arrays.",
       "example": "const difference = (a, b) => a.filter((item) => !b.includes(item))"
     },
+    "snippet::array-initial": {
+      "id": "snippet::array-initial",
```
```suggestion
    "snippet::array-slice-exclude-last-n": {
      "id": "snippet::array-slice-exclude-last-n",
```

or if you have better ideas. as long as it isn't named after one mapping

PR #334, opened 2026-02-25
feat: add `remark-include` and etc to replacements

sergey-s-betke:


gameroman:
I am against this one as suggested replacement has more dependencies than original packages

cc @43081j

PR #344, opened 2026-02-26
feat: add `crc-32` to replacements

### LINKED ISSUE #290: `crc-32` and `buffer-crc32`
```js
import crc32 from "crc-32";

const checksum = crc32.buf(buffer) >>> 0;
```

and

```js
import crc32 from "buffer-crc32";

const checksum = crc32.unsigned(buffer)
```

can be replaced with

```js
import { crc32 } from 'node:zlib';

const checksum = crc32(buffer) >>> 0;
```

gameroman:
Closes #290

### DIFF
--- a/manifests/native.json
+++ b/manifests/native.json
@@ -1337,6 +1337,12 @@
         "compatKey": "api.queueMicrotask"
       }
     },
+    "safe-buffer": {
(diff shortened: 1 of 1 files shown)


43081j:
looks like you need to `npm run sort-manifests`

PR #352, opened 2026-02-26
feat: add `escape-string-regexp` to replacements

### LINKED ISSUE #351: `escape-string-regexp`
escape-string-regexp has 174M weekly downloads. v5 is ESM-only, causing ERR_REQUIRE_ESM for CommonJS consumers. Most projects are stuck on v4 because of this.

This could go in micro-utilities since the core logic is one line:

```js
const escapeStringRegexp = (s) => s.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```

If a packaged alternative is preferred, [tiny-escape](https://www.npmjs.com/package/tiny-escape) wraps this with TypeScript types and ships ESM + CJS. 192 bytes gzipped, zero deps.

[npmgraph comparison](https://npmgraph.js.org/?q=tiny-escape)

gameroman:
Closes #351

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -234,12 +234,6 @@
       "description": "You can create a copy of a regular expression using the `RegExp` constructor.",
       "example": "const copyRegExp = (regexp) => new RegExp(regexp);"
     },
-    "snippet::regexp-escape": {
-      "id": "snippet::regexp-escape",
-      "type": "simple",
-      "description": "You can escape a regular expression using `String.prototype.replace`.",
-      "example": "const escapeStringRegexp = (string) => string.replace(/[|\\\\{}()[\\]^$+*?.]/g, '\\\\$&').replace(/-/g, '\\\\x2d');"
-    },
     "snippet::set-tostringtag": {
       "id": "snippet::set-tostringtag",
       "type": "simple",


manifests/micro-utilities.json
```diff
@@ -222,6 +234,12 @@
       "description": "You can create a copy of a regular expression using the `RegExp` constructor.",
       "example": "const copyRegExp = (regexp) => new RegExp(regexp);"
     },
+    "snippet::regexp-escape": {
+      "id": "snippet::regexp-escape",
+      "type": "simple",
+      "description": "You can escape a regular expression using `String.prototype.replace`.",
+      "example": "const escapeStringRegexp = (string) => string.replace(/[|\\\\{}()[\\]^$+*?.]/g, '\\\\$&').replace(/-/g, '\\\\x2d');"
```
43081j: @bluwy do you have any thoughts on this kind of suggestion? maybe it is too complex to justify recommending people inline it every time. we could just recommend `RegExp.escape` and not recommend anything below node 24 alternatively (i.e. don't flag it at all below 24)

bluwy: In the past I usually copy from mdn (but [they've removed it](https://github.com/mdn/content/commit/a73295d4344aeab38c67262717d0dda8b3b9f0c5#diff-1eaa1a20a76245fe5a7497412d89153d61bf998b1001456485476fd51467b882)) which is a bit shorter. Not sure what the \x2d is about.

If it matters I guess it does make sense to not recommend inlining the snippet then, node 22 with EOL in about a year which isn't too far.

gameroman: `\x2d` is because of [regex unicode strict grammar](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/unicode)

gameroman: I guess we can keep just the `RegExp.escape`

43081j: for now lets remove this snippet and rely on the engines constraint instead

this will only be suggested in engines which have `RegExp.escape` then

PR #354, opened 2026-02-26
feat: add `resolve` package to preferred manifest

### LINKED ISSUE #340: `resolve`
v1 is unmaintained https://npmgraph.js.org/?q=resolve
v2 adds tons of unnecessary dependencies https://npmgraph.js.org/?q=resolve@2.0.0-next.6

Can sometimes use `import.meta.resolve` or `createRequire`.

Netlify is moving away from it using `oxc-resolver`: https://github.com/netlify/build/pull/6955

gameroman:
Closes #340

### DIFF
--- a/manifests/preferred.json
+++ b/manifests/preferred.json
@@ -2988,6 +2988,10 @@
     "import.meta.resolve": {
       "id": "import.meta.resolve",
       "type": "native",
+      "webFeatureId": {
+        "featureId": "js-modules",
+        "compatKey": "javascript.operators.import_meta.resolve"
+      },
       "url": {
         "type": "mdn",
         "id": "Web/JavaScript/Reference/Operators/import.meta/resolve"


docs/modules/resolve.md
```diff
@@ -0,0 +1,35 @@
...
+
+Example:
+
+```ts
+const path = import.meta.resolve("my-module");
+```
+
+## `exsolve`
+
+[`exsolve`](https://www.npmjs.com/package/fast-querystring)
```
43081j: wrong URL here it seems

43081j:
you'll also need to add it to the preferred manifest and run format

manifests/preferred.json
```diff
@@ -2434,6 +2434,12 @@
       "replacements": ["fetch", "ofetch", "ky"],
       "url": {"type": "e18e", "id": "fetch"}
     },
+    "resolve": {
+      "type": "module",
+      "moduleName": "resolve",
+      "replacements": ["exsolve", "oxc-resolver"],
```
43081j: isn't this meant to contain import.meta.resolve too?

manifests/preferred.json
```diff
@@ -2948,6 +2960,15 @@
       "url": {"type": "e18e", "id": "node-telegram-bot-api"},
       "replacementModule": "grammy"
     },
+    "import.meta.resolve": {
+      "id": "import.meta.resolve",
+      "type": "native",
+      "description": "import.meta.resolve",
```
43081j: It doesn't need a description. If possible, it should have a web feature id

gameroman: I think I did this based off https://github.com/gameroman/module-replacements/blob/a7b4f6c8564866e5e7cfe366c248b5b156497db0/manifests/preferred.json#L3213-L3221

I will update both

gameroman: Actually @43081j there is no web feature id for `import.meta` from what I see

43081j: It's in the js-modules feature I think

gameroman: > It's in the js-modules feature I think

I assumed all of that comes from https://github.com/web-platform-dx/web-features ?

If that is so, I think it [isn't there](https://github.com/search?q=repo%3Aweb-platform-dx%2Fweb-features%20%22import.meta%22&type=code)

43081j: it is in js-modules:

https://github.com/web-platform-dx/web-features/blob/d01dd9014d5eb5929922dab0a685443f09bcbdf4/features/js-modules.yml.dist#L100-L111

using github search isn't very useful here. you should just load it locally and look at the compat keys by eye as there are not many per feature

gameroman: Oh I see, thank you!

43081j (approved):

PR #357, opened 2026-02-27
feat: add `buffer-equal-constant-time` to replacements

### LINKED ISSUE #356: `buffer-equal-constant-time`
Can be replaced with

```js
import * as crypto from 'crypto'

const a = Buffer.from('a')
const b = Buffer.from('b')

(a.length === b.length && crypto.timingSafeEqual(a, b))
```

gameroman:
Closes #356

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -318,6 +318,11 @@
       "moduleName": "base64-js",
       "replacements": ["snippet::base64"]
     },
+    "buffer-equal-constant-time": {
+      "type": "module",
+      "moduleName": "buffer-equal-constant-time",
+      "replacements": ["snippet::buffer-equal-constant-time"]
+    },
     "call-bind": {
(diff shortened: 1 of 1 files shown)


manifests/micro-utilities.json
```diff
@@ -72,6 +72,12 @@
       "description": "Every modern runtime provides a way to convert byte array to and from base64.",
       "example": "// From base64 to Uint8Array\nconst bytes = Uint8Array.fromBase64(base64)\n// From Uint8Array to base64\nconst base64 = bytes.toBase64()"
     },
+    "snippet::buffer-equal-constant-time": {
```
43081j, line 75: did you forget to add the mapping for this?

gameroman: Looks like I did, indeed

43081j (approved):

PR #366, opened 2026-03-01
feat: add `is-arguments` to replacements

### LINKED ISSUE #360: `is-arguments`
`is-arguments` can be replaced with

```js
const isArguments = (val) => val?.[Symbol.toStringTag] === undefined && Object.prototype.toString.call(val) === '[object Arguments]';
```

gameroman:
Closes #360

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -93,7 +93,7 @@
     "snippet::is-arguments": {
       "id": "snippet::is-arguments",
       "type": "simple",
-      "description": "You can use `Symbol.toStringTag` together with `Object.prototype.toString.call(obj) === \"[object Arguments]\"`",
+      "description": "You can use `Object.prototype.toString.call(obj) === \"[object Arguments]\"`",
       "example": "const isArguments = (val) => Object.prototype.toString.call(val) === \"[object Arguments]\";"
     },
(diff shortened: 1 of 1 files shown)


manifests/micro-utilities.json
```diff
@@ -90,6 +90,12 @@
       "description": "You can use the includes method on the process.argv array to check if a flag is present.",
       "example": "process.argv.includes('--flag')"
     },
+    "snippet::is-arguments": {
+      "id": "snippet::is-arguments",
+      "type": "simple",
+      "description": "You can use `Symbol.toStringTag` together with `Object.prototype.toString.call(obj) === \"[object Arguments]\"`",
```
43081j: Presumably the string tag check is because the source module does that too?

I feel like it's unnecessary in most cases, so maybe we should simplify and leave it out 

Tell me what you think before changing it

gameroman: To check against something like
```js
const fakeArgs = {
  [Symbol.toStringTag]: 'Arguments'
};
```

Can be removed probably

43081j: i think our suggestion should match the common use case, and people who do have a custom string tag can figure it out through docs

43081j (approved):

Issue #247, opened 2025-09-20
More entries to the manifests


outslept:
1. [shebang-regex](https://www.npmjs.com/package/shebang-regex)

the code is

```js
const shebangRegex = /^#!(.*)/;
```

the biggest consumer is cross-spawn through a transitive dep (I've opened a PR there to inline their tiny deps, but I think this is still a good option to add to microutillities manifest)

2.  [once](https://www.npmjs.com/package/once)

I think we can do something like and adapt based on the use case?

```js
export function once(fn) {
  if (typeof fn !== 'function') throw new TypeError('Expected a function');

  let called = false;
  let value;

  return function (...args) {
    if (called) return value;
    called = true;
    value = fn.apply(this, args);
    return value;
  };
}
```

3. [is-potential-custom-element-name](https://www.npmjs.com/package/is-potential-custom-element-name)

```js
const isCustomElementName = name => /^[a-z](?:[\.0-9_a-z\xB7\xC0-\xD6\xD8-\xF6\xF8-\u037D\u037F-\u1FFF\u200C\u200D\u203F\u2040\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD]|[\uD800-\uDB7F][\uDC00-\uDFFF])*-(?:[\x2D\.0-9_a-z\xB7\xC0-\xD6\xD8-\xF6\xF8-\u037D\u037F-\u1FFF\u200C\u200D\u203F\u2040\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD]|[\uD800-\uDB7F][\uDC00-\uDFFF])*$/.test(name);
```

4. sinon -> tinyspy?

for basic stuff tinyspy is good enough as is, for network you should be able to take nock/MSW, take @sinonjs/fake-timers for timers, for module mocking either rely on your test runner or use esmock/mock-require and most folks should be good to go? (idk what to do with Sinon sandboxes though)

43081j:
for the CE name module im not too concerned since its a non-trivial regex that was written to follow the web components standard

sinon usually can be replaced by tinyspy, you are right. i'd like some more opinions on it though since its a big non-drop-in replacement (i.e. it isn't trivial to migrate)

for `once` - i suspect many consumers don't even need it. would be curious what _real people_ install it (its probably another deep dependency situation)

outslept:
once:

```
[...]

PR #370, opened 2026-03-01
add `shebang-regex` to replacements

gameroman:
See #247

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -237,8 +237,8 @@
     "snippet::shebang-regex": {
       "id": "snippet::shebang-regex",
       "type": "simple",
-      "description": "You can use `/^#!(.*)/` regex.",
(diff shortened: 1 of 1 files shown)


manifests/micro-utilities.json
```diff
@@ -234,6 +234,12 @@
       "description": "You can set the `toStringTag` of an object using `Object.defineProperty`.",
       "example": "const setToStringTag = (target, value) => Object.defineProperty(target, Symbol.toStringTag, { value, configurable: true });"
     },
+    "snippet::shebang-regex": {
+      "id": "snippet::shebang-regex",
+      "type": "simple",
+      "description": "You can use `/^#!(.*)/` regex.",
```
43081j: It should use `+` rather than `*` since shebang-regex is actually wrong. A valid shebang needs at least one character

PR #386, opened 2026-03-02
feat: add `is-in-ssh` to replacements

### LINKED ISSUE #379: `is-in-ssh`
```js
const isInSsh = Boolean(process.env.SSH_CONNECTION || process.env.SSH_CLIENT || process.env.SSH_TTY);
```

gameroman:
Closes #379

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -135,8 +135,8 @@
     "snippet::is-in-ssh": {
       "id": "snippet::is-in-ssh",
       "type": "simple",
-      "description": "You can check if the current environment is SSH by checking if the `SSH_CONNECTION`, `SSH_CLIENT` or `SSH_TTY` environment variable is set.",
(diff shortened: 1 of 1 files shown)


manifests/micro-utilities.json
```diff
@@ -132,6 +132,12 @@
       "description": "You can use the modulo operator to check if a number is even.",
       "example": "(n % 2) === 0"
     },
+    "snippet::is-in-ssh": {
+      "id": "snippet::is-in-ssh",
+      "type": "simple",
+      "description": "You can check if the current environment is SSH by checking if the `SSH_CONNECTION`, `SSH_CLIENT` or `SSH_TTY` environment variable is set.",
```
43081j: Are we sure we need to check each one? It would be worth doing some reading on this as it may be the case that we only need to check for SSH_CONNECTION

gameroman: Seems like legacy reasons mostly. So yes, we only need `SSH_CONNECTION` now

gameroman: Looking at downloads, looks like only one package has it as a dependency

https://dependents.dev/?package=is-in-ssh

PR #388, opened 2026-03-02
feat: add `get-own-enumerable-keys` to replacements

### LINKED ISSUE #382: `get-own-enumerable-keys`
[`get-own-enumerable-keys`](https://npmx.dev/package/get-own-enumerable-keys)

gameroman:
Closes #382

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -90,6 +90,12 @@
       "description": "Every modern runtime provides a way to get the iterator function through the Symbol.iterator symbol.",
       "example": "const iterator = obj[Symbol.iterator]?.()"
     },
+    "snippet::get-own-enumerable-keys": {
+      "id": "snippet::get-own-enumerable-keys",
+      "type": "simple",
+      "description": "You can get own enumerable keys using `Object.keys` and `Object.getOwnPropertySymbols` with `Object.prototype.propertyIsEnumerable`.",
+      "example": "Object.keys(object).concat(Object.getOwnPropertySymbols(object).filter(key => Object.prototype.propertyIsEnumerable.call(object, key)))"
+    },
     "snippet::has-argv": {
       "id": "snippet::has-argv",
       "type": "simple",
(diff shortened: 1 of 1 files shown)


manifests/micro-utilities.json
```diff
@@ -90,6 +90,12 @@
       "description": "Every modern runtime provides a way to get the iterator function through the Symbol.iterator symbol.",
       "example": "const iterator = obj[Symbol.iterator]?.()"
     },
+    "snippet::get-own-enumerable-keys": {
+      "id": "snippet::get-own-enumerable-keys",
+      "type": "simple",
+      "description": "You can get own enumerable keys using `Object.keys` and `Object.getOwnPropertySymbols` with `Object.prototype.propertyIsEnumerable`.",
```
43081j, line 96: Doesn't `Object.keys` already return enumerable own keys? So maybe we don't need to complicate this with getOwnPropertySymbols?

gameroman: > Doesn't `Object.keys` already return enumerable own keys? So maybe we don't need to complicate this with getOwnPropertySymbols?

It doesn't return symbols for legacy reasons

gameroman: I've just double checked and this package is only used by one package

https://dependents.dev/?package=get-own-enumerable-keys

| # | Downloads/month | Traffic | Version | Package |
|---|-----------------|---------|---------|---------|
| 1 | 90.98M | 157.39 GB | ^1.0.0 | [stringify-object](https://npmx.dev/stringify-object) |

43081j: we could leave it out if you like and just drop `stringify-object`

gameroman: Drop `stringify-object`, as in add it to replacements?

43081j: as an ecosystem effort rather than in here. its basically a serialisation strategy, so i doubt many of its users _really_ need it and could just use JSON.

something for us to look into

lets close this for now and look into who uses stringify-object

gameroman: I'll open an ecosystem issue

gameroman: e18e/ecosystem-issues#239

PR #392, opened 2026-03-03
move `buffer-equal-constant-time` to native replacements

### LINKED ISSUE #391: Update buffer comparison to be a native replacement
> 
> - move it to `native.json`
> - change the replacement to be `crypto.subtle.timingSafeEqual`, set the right node ID/URL, etc
> - add an e18e doc URL to the mapping
> - add the doc to show these 2 examples
> 

 _Originally posted by @43081j in [#356](https://github.com/es-tooling/module-replacements/issues/356#issuecomment-3990493739)_

This is in regards to the buffer equality check suggestion

gameroman:
Closes #391

### DIFF
--- a/docs/modules/buffer-equal-constant-time.md
+++ b/docs/modules/buffer-equal-constant-time.md
@@ -8,36 +8,7 @@ description: Native Node.js alternatives to the buffer-equal-constant-time packa
 
 You can use the [`timingSafeEqual`](https://nodejs.org/api/crypto.html#cryptotimingsafeequala-b) function from the `node:crypto` module.
 
-If it is known that `buf1` and `buf2` have same size:
-
-```ts
-import { Buffer } from 'node:buffer'
-import bufferEqual from 'buffer-equal-constant-time' // [!code --]
-import * as crypto from 'node:crypto' // [!code ++]
-
-const buf1 = Buffer.from('303')
-const buf2 = Buffer.from('303')
-
-bufferEqual(buf1, buf2) // [!code --]
-crypto.timingSafeEqual(buf1, buf2) // [!code ++]
-```
-
-If size of `bufSecret` is publicly known:
-
-```ts
-import { Buffer } from 'node:buffer'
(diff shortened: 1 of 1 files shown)


docs/modules/buffer-equal-constant-time.md
```diff
@@ -0,0 +1,54 @@
...
+import bufferEqual from 'buffer-equal-constant-time' // [!code --]
+import * as crypto from 'node:crypto' // [!code ++]
+
+const bufUser = Buffer.from('303')
+const bufSecret = Buffer.from('303')
+
+bufferEqual(bufUser, bufSecret) // [!code --]
+bufUser.length === bufSecret.length // [!code ++]
+  ? crypto.timingSafeEqual(bufUser, bufSecret) // [!code ++]
+  : !crypto.timingSafeEqual(bufUser, bufSecret) // [!code ++]
```
43081j: This is meant to be bufUser in both parameters

docs/modules/buffer-equal-constant-time.md
```diff
@@ -0,0 +1,54 @@
...
+
+const bufUser = Buffer.from('303')
+const bufSecret = Buffer.from('303')
+
+bufferEqual(bufUser, bufSecret) // [!code --]
+bufUser.length === bufSecret.length && // [!code ++]
+  crypto.timingSafeEqual(bufUser, bufSecret) // [!code ++]
+```
+
+If size of `bufSecret` is not publicly known:
```
43081j: I'm not sure we need to distinguish this. We should probably just list this one and get rid of the one above that is "publicly known"

gameroman: @imcotton what do you think?

imcotton: I also think this is a rare use case. By the time you reach for safety, it is a net negative and error-prone by having presumption of "publicly known" or "same size".

PR #396, opened 2026-03-07
feat: add `through` package to replacements

### LINKED ISSUE #47: replacement for through and @ljharb/through
We should add them to be a replacement candidate, as it can be replaced with native WriteableStream

https://github.com/SBoudrias/Inquirer.js/pull/1439/ 
https://github.com/feross/hostile/pull/55

@SBoudrias
@kibertoad

gameroman:
Closes #47

### DIFF
--- a/docs/modules/through.md
+++ b/docs/modules/through.md
@@ -6,6 +6,7 @@ description: Modern alternatives to the through package
 
 ## `stream.Writable` (native, Node.js)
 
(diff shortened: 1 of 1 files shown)


43081j:
given its in the "preferred" manifest, i think that's ok as this is specifically the "opinionated" manifest.

people who prefer the syntax or need to support older node can turn this one off

PR #397, opened 2026-03-07
feat: add `is-generator-function` to replacements

### LINKED ISSUE #122: `is-generator-function` native replacement
https://github.com/inspect-js/is-generator-function

There is a one-liner replacement:

```ts
const isGenerator = typeof fn === 'function' && Object.getPrototypeof(fn) === (function* () {}.constructor);
```

Though you probably want to cache the generator prototype (JS should make it a global, IMO).

`is-generator-function` depends on `has-tostringtag` (which depends on a third) so replacing this could eliminate up to three dependencies.

gameroman:
Closes #122

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -135,7 +135,7 @@
     "snippet::is-generator-function": {
       "id": "snippet::is-generator-function",
       "type": "simple",
-      "description": "You can use `typeof` and `Object.prototype.toString.call` to check if it's a generator function",
+      "description": "You can use `typeof` and `Object.prototype.toString.call` to check if a value is a generator function",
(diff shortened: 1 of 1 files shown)


manifests/micro-utilities.json
```diff
@@ -132,6 +132,12 @@
       "description": "You can use the modulo operator to check if a number is even.",
       "example": "(n % 2) === 0"
     },
+    "snippet::is-generator-function": {
+      "id": "snippet::is-generator-function",
+      "type": "simple",
+      "description": "You can use `typeof` and `Object.prototype.toString.call` to check if it's a generator function",
```
```suggestion
      "description": "You can use `typeof` and `Object.prototype.toString.call` to check if a value is a generator function",
```

PR #405, opened 2026-03-07
feat: add `teeny-request` and `retry-request` to replacements

gameroman:


gameroman:
> I'm not sure retry-request is replaced by fetch. It exists to provide a retry strategy on top of request (or on top of fetch in modern times).
> 
> We should suggest an alternative lightweight package that adds a retry strategy to fetch I think

Updated to only recommend `ofetch` and `ky`. Should be fine now I think

manifests/preferred.json
```diff
@@ -2410,6 +2410,12 @@
       "replacements": ["fetch", "ky", "ofetch"],
       "url": {"type": "e18e", "id": "fetch"}
     },
+    "retry-request": {
+      "type": "module",
+      "moduleName": "retry-request",
+      "replacements": ["ky", "ofetch"],
```
43081j:
```suggestion
      "replacements": ["ofetch", "ky"],
```

lets prefer ofetch since the alpha (soon to be stable) shrinks a lot

same for below too

gameroman: I thought this was alphabetical as everywhere. Should I do this everywhere then?

43081j: it is meant to be ordered by which one we prefer (since v3).

you're just submitting new entries before we've had chance to document any of this stuff since v3 is brand new

PR #413, opened 2026-03-12
feat: add `react-helmet` to preferred manifest

### LINKED ISSUE #185: React Helmet
React Helmet hasn't been updated in 4 years https://www.npmjs.com/package/react-helmet compared to the alternative package [React Helmet async](https://www.npmjs.com/package/react-helmet-async) which was last updated 7 months ago. There's also console errors about using unsafe methods when using React Helmet https://github.com/nfl/react-helmet/issues/694

gameroman:
Closes #185

### DIFF
--- a/manifests/preferred.json
+++ b/manifests/preferred.json
@@ -2437,7 +2437,7 @@
     "react-helmet": {
       "type": "module",
       "moduleName": "react-helmet",
-      "replacements": ["react-helmet-async"],
+      "replacements": ["react:metadata", "react-helmet-async"],
       "url": {"type": "e18e", "id": "react-helmet"}
     },
     "read-package-up": {
@@ -3190,6 +3190,12 @@
       "type": "documented",
       "url": {"type": "e18e", "id": "react-helmet"},
       "replacementModule": "react-helmet-async"
+    },
+    "react:metadata": {
+      "id": "react:metadata",
+      "type": "documented",
+      "url": {"type": "e18e", "id": "react-helmet"},
+      "replacementModule": "react"
     },
     "readline.createInterface": {
       "id": "readline.createInterface",


docs/modules/react-helmet.md
```diff
@@ -0,0 +1,47 @@
...
+[`react-helmet-async`](https://github.com/staylor/react-helmet-async) is a fork of `react-helmet`.
+
+Example:
+
+```jsx
+import { Helmet } from 'react-helmet' // [!code --]
+import { Helmet, HelmetProvider } from 'react-helmet-async' // [!code ++]
+
+const app = (
+  <HelmetProvider>
```
43081j: this was added, no? i think you missed the diff

gameroman: Yeah, I was unsure if this comment works the same way in jsx

manifests/preferred.json
```diff
@@ -2404,6 +2404,12 @@
       "replacements": ["nanoid"],
       "url": {"type": "e18e", "id": "shortid"}
     },
+    "react-helmet": {
+      "type": "module",
+      "moduleName": "react-helmet",
+      "replacements": ["react-helmet-async"],
```
joaopedrodcf: How should we also tell that we can actually just use react without a module maybe this part is important for when it appears in npmx

gameroman: Not sure there is a way to do that at the moment

43081j: really there's two replacements, so this array isn't quite right.

the first replacement should probably be "react" i guess.

we can figure out nicer ways to present that and infer that it means "no dependency needed" in future

gameroman: I was thinking in future we could add something like `:remove:` or `:upstreamed:` which would that functionality exists in the package for which this is used or something like that

43081j: for now we can just add react.

since you are in fact replacing the thing you use with something the `"react"` module provides

gameroman: Makes sense

PR #416, opened 2026-03-13
feat: add `define-property` to native manifest

gameroman:


gameroman:
I used `Reflect.defineProperty` as the library mentioned it. Should I add both replacements to the array or just keep `Object.defineProperty`?

joaopedrodcf (approved):
Nice job

PR #439, opened 2026-03-15
feat: add `event-stream` to preferred manifest

### LINKED ISSUE #409: suggest replacing `event-stream` with native node.js functionality
# [`event-stream`](https://npmx.dev/package/event-stream)

## `es.pipeline()` replacement

```ts
import es from 'event-stream' // [!code --]
import { pipeline } from 'node:stream/promises' // [!code ++]

es.pipeline(s1, s2, s3) // [!code --]
await pipeline(s1, s2, s3) // [!code ++]
```

## `es.split()` replacement

```ts
import es from 'event-stream' // [!code --]
import { createReadStream } from 'node:fs' // [!code ++]
import { createInterface } from 'node:readline' // [!code ++]

createReadStream('file.txt').pipe(es.split()) // [!code --]
const rl = createInterface({ input: createReadStream('file.txt') }) // [!code ++]
for await (const line of rl) { /* ... */ } // [!code ++]
```

## `es.map()`

```ts
import es from 'event-stream' // [!code --]
import { Readable } from 'node:stream' // [!code ++]

es.map((data, cb) => cb(null, data.toUpperCase())) // [!code --]
Readable.from(source).map(data => data.toUpperCase()) // [!code ++]
```

## `es.merge()`

```ts
import es from 'event-stream' // [!code --]
import { PassThrough } from 'node:stream' // [!code ++]

es.merge([s1, s2]) // [!code --]
const combined = new PassThrough() // [!code ++]
s1.pipe(combined, { end: false }) // [!code ++]
s2.pipe(combined, { end: false }) // [!code ++]
```

gameroman:
Closes #409

### DIFF
--- a/docs/modules/README.md
+++ b/docs/modules/README.md
@@ -41,6 +41,7 @@ ESLint plugin.
 - [`eslint-plugin-node`](./eslint-plugin-node.md)
 - [`eslint-plugin-react`](./eslint-plugin-react.md)
 - [`eslint-plugin-vitest`](./eslint-plugin-vitest.md)
+- [`event-stream`](./event-stream.md)
 - [`execa`](./execa.md)
 - [`ezspawn`](./ez-spawn.md)
 - [`faker`](./faker.md)


docs/modules/event-stream.md
```diff
@@ -0,0 +1,51 @@
+---
+description: Modern alternatives to the event-stream package
+---
+
+# Replacements for `event-stream`
+
+## `es.pipeline()` replacement
```
43081j, line 7: the 2nd heading should be:

```md
## pipelines (native, Node.js)
```

then within that, these 3 sections

edit: ah i see actually none of these are really related. that is a pain. 

maybe just:

```md
## `node:stream` and readline (native, Node.js)
```

there has to be something as these 3 headings are not the replacements, the replacement is `node:stream`

gameroman: I was just doing it the same way I did it with https://e18e.dev/docs/replacements/get-stream

gameroman: Although there the titles are more generic

43081j: in both it should really have an outer heading but im not sure what.

for now lets just  leave this and we can tackle both in a future pr

PR #450, opened 2026-03-16
feat: add `wellknown` to preferred manifest

### LINKED ISSUE #432: `betterknown`
I think replacing https://npmx.dev/package/wellknown with https://npmx.dev/package/betterknown is a good recommendation.

https://npmx.dev/compare?packages=wellknown,betterknown

Both packages are fairly stable in what they do but `wellknown` is was [not updated in 10 year](https://github.com/mapbox/wellknown) and `betterknown` is a modern and maintained rewrite that has be present for a few years now.

gameroman:
Closes #432

### DIFF
--- a/docs/modules/wellknown.md
+++ b/docs/modules/wellknown.md
@@ -1,5 +1,5 @@
 ---
-description: Modern replacements for the unmaintained wellknown package, a WKT parser and stringifier
+description: Modern replacements for the wellknown package, a WKT parser and stringifier
 ---
 
 # Replacements for `wellknown`


docs/modules/wellknown.md
```diff
@@ -0,0 +1,32 @@
+---
+description: Modern replacements for the unmaintained wellknown package, a WKT parser and stringifier
+---
+
+# Replacements for `wellknown`
+
+## `betterknown`
+
+[`betterknown`](https://github.com/placemark/betterknown) is a maintained alternative for `wellknown` package.
```
joaopedrodcf, line 9: Should we say that is from the same maintainer ? But as the other one belonged to mapbox and he left it started being unmantained

docs/modules/wellknown.md
```diff
@@ -0,0 +1,32 @@
+---
+description: Modern replacements for the unmaintained wellknown package, a WKT parser and stringifier
```
43081j:
```suggestion
description: Modern replacements for the wellknown package, a WKT parser and stringifier
```

docs/modules/wellknown.md
43081j: i think if they still work at mapbox that probably makes sense, but im unsure if thats the case. so i dont think we should try give it authenticity by linking it to mapbox in that case

ill merge this as is for now but we can always tweak it later

PR #451, opened 2026-03-16
docs: move `inherits` to preferred manifest

### LINKED ISSUE #147: `inherits` entry is not dev friendly
The `inherits` package was added with #103 to solve #99 by @justinfagnani

I politely disagree with the whole remark, that you should just use "class syntax". Nobody with the right mind would imho rewrite everything to shave off 4 kb.  Also just to claim that inherits is considered legacy, does not mean that it is deprecated. It just means that it will stay but will not be actively developed.

It would be closer to reality to recommend what inherits actually does by directly modifying the prototype of the Target with the Base

So instead of:

```
inherits(Target, Base)
```

you just do:

 ```js
 Target.prototype = Object.create(Base.prototype, {
    constructor: { value: Target, enumerable: false, writable: true, configurable: true }
  })
```

So you dont need to rewrite the functions to classes, just to remove inherits.

### PR DESCRIPTION
**Pull request by gameroman**
Closes #147

### DIFF
--- a/manifests/preferred.json
+++ b/manifests/preferred.json
@@ -327,7 +327,8 @@
     "inherits": {
       "type": "module",
       "moduleName": "inherits",
-      "replacements": ["extends", "util.inherits"]
+      "replacements": ["extends", "util.inherits"],
+      "url": {"type": "e18e", "id": "inherits"}
     },
     "invariant": {
       "type": "module",

### CODE REVIEW

File: manifests/native.json [hunk 1]
```diff
@@ -1531,12 +1537,6 @@
         "id": "Web/JavaScript/Reference/Global_Objects/TypedArray"
       }
     },
-    "inherits": {
-      "id": "inherits",
-      "type": "removal",
-      "description": "You can just use the `class` syntax.",
```
43081j, line 1537: we should probably just keep `inherits` since all other removals work this way (i.e. every removal and its mapping have the same id).

we just need to reword the description and set the URL. we should also mention `Object.create`

gameroman: It's not exactly a 'removal', I think it should be either 'documented' or 'native'. 'removal' is more for things that can be removed without adding anything in return

gameroman: I think making it 'documented' would be better so we could better explain how to move from `inherits` to something else

43081j: you are right, its not a removal. it is native though - just there are two possibly syntaxes to replace it.

because of that, sure lets put it in preferred manifest for now as a documented replacement.

very simple example in the doc of `extends` and `Object.create`

all done then

43081j (approved):

PR #453, opened 2026-03-17
feat: add `temp` package to preferred manifest

### LINKED ISSUE #339: suggest replacing `temp` with native node.js `mkdtemp`
[`temp`](https://npmx.dev/temp)

[`mkdtemp`](https://nodejs.org/api/fs.html#fspromisesmkdtempprefix-options)

gameroman:
Closes #339

### DIFF
--- a/docs/modules/tempy.md
+++ b/docs/modules/tempy.md
@@ -11,16 +11,6 @@ Node.js has the [`fs.mkdtemp`](https://nodejs.org/api/fs.html#fsmkdtempprefix-op
 Example:
 
 ```ts
-import { temporaryDirectory } from 'tempy' // [!code --]
-import { mkdtemp, realpath } from 'node:fs/promises' // [!code ++]
-import { join } from 'node:path' // [!code ++]
-import { tmpdir } from 'node:os' // [!code ++]
-
-const tempDirPath = temporaryDirectory() // [!code --]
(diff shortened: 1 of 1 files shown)


gameroman:
Should I then instead of diffs make it examples without diffs or should I keep the diffs for `tempy`?

docs/modules/tempy.md
```diff
@@ -16,8 +16,36 @@ import { mkdtemp, realpath } from 'node:fs/promises' // [!code ++]
 import { join } from 'node:path' // [!code ++]
 import { tmpdir } from 'node:os' // [!code ++]
 
-const tempDir = temporaryDirectory() // [!code --]
-const tempDir = await mkdtemp(join(await realpath(tmpdir()), 'foo-')) // [!code ++]
+const tempDirPath = temporaryDirectory() // [!code --]
+const tempDirPath = await mkdtemp(join(await realpath(tmpdir()), 'foo-')) // [!code ++]
+```
+
+```ts
```
43081j, line 13: this code block needs some context. a description or something. what is it showing us that the earlier one doesn't show?

gameroman: It's that one is showing diff for `tempy` and the second one for `temp`

43081j: currently its just a standalone code block with no explanation. we should add some text before it, or not have it.

we haven't yet decided how to handle multiple target modules in one doc. this is the first time. so im not sure if we should bother as it feels like a discussion is needed

PR #454, opened 2026-03-17
feat: add `collection-map` to native manifest with doc

### LINKED ISSUE #377: `collection-map`
[`collection-map`](https://npmx.dev/package/collection-map)

gameroman:
Closes #377

### DIFF
--- a/docs/modules/collection-map.md
+++ b/docs/modules/collection-map.md
@@ -6,17 +6,24 @@ description: Native alternatives to the collection-map package
 
 ## Objects
 
+<!-- prettier-ignore -->
 ```js
 const map = require('collection-map') // [!code --]
 const data = { a: 'foo', b: 'bar', c: 'baz' }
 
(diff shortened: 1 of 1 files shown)


docs/modules/collection-map.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Objects
+
+```js
+const map = require('collection-map') // [!code --]
+const data = { a: 'foo', b: 'bar', c: 'baz' }
+
+// Mapping values
+const res = map(data, (item) => item) // [!code --]
+const res = Object.values(data) // [!code ++]
```
43081j: this isn't a very useful example.

it is how you would migrate `map(data, (item) => item)` indeed, but thats because the initial example is showing the syntax rather than anything meaningful. if you're going to show the before/after, you need to pick a more meaningful one that doesn't get lost in conversion. one that has a map still in the end

PR #471, opened 2026-03-22
feat: switch to ESM-only and import attributes

43081j:
This switches to ESM only, which means:

- We drop `tshy` and just use TypeScript directly
- Use import attributes (i.e. import the JSON directly using `import`)

This does mean we require a newer Node, one that can `require(esm)` and
use import attributes (specifically JSON).

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
         "default": "./dist/main.js"
       }
     },
+    "./manifests/*.json": "./manifests/*.json",
     "./package.json": "./package.json"
   },
   "types": "./dist/main.d.ts",


package.json
```diff
@@ -51,26 +42,18 @@
...
   "exports": {
     ".": {
       "import": {
-        "types": "./dist/esm/main.d.ts",
-        "default": "./dist/esm/main.js"
-      },
-      "require": {
-        "types": "./dist/commonjs/main.d.ts",
-        "default": "./dist/commonjs/main.js"
+        "types": "./dist/main.d.ts",
```
joaopedrodcf, line 51: What do you think of us also adding to the package json the node version minimum for the package ?

43081j: i think we should but maybe lets do it in a follow up once we agree on the right one

PR #472, opened 2026-03-23
docs: add contribution guide

43081j:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,7 +64,7 @@ When using the `native` type:
 - The mapping generally does not need a `url`
 - The replacement should have a `url` to its external documentation (e.g. MDN or Node docs)
 - The replacement should have a `webFeatureId` if it is a web standard (these come from the [`web-features` dataset](https://github.com/web-platform-dx/web-features/tree/main/features))
-- The replacement should have a `nodeFeatureId` if it is a Node built-in (these are of the form `{moduleName[, exportName]}`. For example, `fs` would be `{"moduleName": "fs"}`, while `fs.readFile` would be `{"moduleName": "fs", "exportName": "readFile"}`)
+- The replacement should have a `nodeFeatureId` if it is a Node built-in (these are of the form `{ moduleName: string, exportName?: string }`. For example, `fs` would be `{"moduleName": "fs"}`, while `fs.readFile` would be `{"moduleName": "fs", "exportName": "readFile"}`)
 
 When using the `simple` type:
 


joaopedrodcf (approved):

CONTRIBUTING.md
```diff
@@ -0,0 +1,110 @@
...
+- The mapping should have a `url` which points to an e18e documentation page (i.e. those in the [modules](https://github.com/es-tooling/module-replacements/tree/main/docs/modules) directory of the repo) which describes the replacement and how to migrate to it.
+- The replacement does not necessarily need a `url` itself
+- The replacement does not need a `description`
+
+When using the `native` type:
+
+- The mapping generally does not need a `url`
+- The replacement should have a `url` to its external documentation (e.g. MDN or Node docs)
+- The replacement should have a `webFeatureId` if it is a web standard (these come from the [`web-features` dataset](https://github.com/web-platform-dx/web-features/tree/main/features))
+- The replacement should have a `nodeFeatureId` if it is a Node built-in (these come from the Node documentation)
```
paoloricciuti: Should we add an example here to make it easier to understand where to get the `nodeFeatureId`?

CONTRIBUTING.md
```diff
@@ -0,0 +1,110 @@
...
+
+When using the `simple` type:
+
+- The replacement should be of the form `snippet::{name}` where `name` is a suitable descriptive name for what the snippet does (e.g. `snippet:is-odd` for a snippet that checks if a number is odd)
+- The replacement description should be a short human-readable description of what the snippet does, and how it replaces the old module
```
paoloricciuti: I remember having some issue with this when moving the old data so maybe we could have an example here?

CONTRIBUTING.md
```diff
@@ -0,0 +1,110 @@
...
+- Replacement keys match their `id` property
+- `webFeatureId` values reference valid features and compat keys from the `web-features` dataset
+- Mapping keys match their `moduleName` for module-type mappings
+- Mappings only reference replacements that exist
+- All documented modules (in `docs/modules/`) are listed in the modules README
+- All documented modules have a corresponding entry in a manifest
+- No module appears in multiple manifests
+- Mappings and replacements are sorted alphabetically
+- No replacement has `engines` set (these are populated automatically at publish time)
+- All e18e URLs in mappings and replacements point to documentation files that exist
```
paoloricciuti: Do we need to keep this list here? It can easily go out of sync with the lint and if the errors from the scripts are readable it would be much easier to run the scripts and get the errors instead of reading this list.

paoloricciuti (commented):
Overall looks good...left a few comments but are mostly minor things

CONTRIBUTING.md
gameroman: I think we can keep part of the list

CONTRIBUTING.md
```diff
@@ -0,0 +1,105 @@
...
+- The mapping should have a `url` which points to an e18e documentation page (i.e. those in the [modules](https://github.com/es-tooling/module-replacements/tree/main/docs/modules) directory of the repo) which describes the replacement and how to migrate to it.
+- The replacement does not necessarily need a `url` itself
+- The replacement does not need a `description`
+
+When using the `native` type:
+
+- The mapping generally does not need a `url`
+- The replacement should have a `url` to its external documentation (e.g. MDN or Node docs)
+- The replacement should have a `webFeatureId` if it is a web standard (these come from the [`web-features` dataset](https://github.com/web-platform-dx/web-features/tree/main/features))
+- The replacement should have a `nodeFeatureId` if it is a Node built-in (these are of the form `{moduleName[, exportName]}`. For example, `fs` would be `{"moduleName": "fs"}`, while `fs.readFile` would be `{"moduleName": "fs", "exportName": "readFile"}`)
```
paoloricciuti: Why not this?

```suggestion
- The replacement should have a `nodeFeatureId` if it is a Node built-in (these are of the form `{ moduleName: string, exportName?: string }`. For example, `fs` would be `{"moduleName": "fs"}`, while `fs.readFile` would be `{"moduleName": "fs", "exportName": "readFile"}`)
```

CONTRIBUTING.md
```diff
@@ -64,12 +64,12 @@ When using the `native` type:
...
 - The replacement should have a `webFeatureId` if it is a web standard (these come from the [`web-features` dataset](https://github.com/web-platform-dx/web-features/tree/main/features))
-- The replacement should have a `nodeFeatureId` if it is a Node built-in (these come from the Node documentation)
+- The replacement should have a `nodeFeatureId` if it is a Node built-in (these are of the form `{moduleName[, exportName]}`. For example, `fs` would be `{"moduleName": "fs"}`, while `fs.readFile` would be `{"moduleName": "fs", "exportName": "readFile"}`)
 
 When using the `simple` type:
 
 - The replacement should be of the form `snippet::{name}` where `name` is a suitable descriptive name for what the snippet does (e.g. `snippet:is-odd` for a snippet that checks if a number is odd)
-- The replacement description should be a short human-readable description of what the snippet does, and how it replaces the old module
+- The replacement description should be a short human-readable description of what the snippet does, and how it replaces the old module (e.g. "You can use arr.at(-1) to get the last element of an array")
```
paoloricciuti, line 72: Mmm I know I'm being pedantic but this doesn't look like the right example because it includes the snippets itself (and the descriptions for those should not include the snippet no?)

What about

```suggestion
- The replacement description should be a short human-readable description of what the snippet does, and how it replaces the old module (e.g. "You can use a combination of filter and includes to calculate the difference between two arrays.")
```

gameroman: Both are fine I think

CONTRIBUTING.md
43081j: no reason

CONTRIBUTING.md
43081j: there is no restriction of a description mentioning code. in this example it makes perfect sense to mention code in the description, because it is the method name. rewording it to avoid mentioning code doesn't make sense.

gameroman (approved):

CONTRIBUTING.md
```diff
@@ -0,0 +1,105 @@
...
+
+When using the `documented` type, you should generally follow these guidelines:
+
+- The mapping should have a `url` which points to an e18e documentation page (i.e. those in the [modules](https://github.com/es-tooling/module-replacements/tree/main/docs/modules) directory of the repo) which describes the replacement and how to migrate to it.
+- The replacement does not necessarily need a `url` itself
+- The replacement does not need a `description`
+
+When using the `native` type:
+
+- The mapping generally does not need a `url`
```
ghostdevv:
```suggestion
- The mapping generally does not need a `url`, but if it does then it's because ...
```

not sure what the `...` should be here

PR #473, opened 2026-03-23
docs: small contribution guide tweaks

ghostdevv:
A few things that I think cleanup some potential questions or confusion

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,8 +63,8 @@ When using the `native` type:
 
 - The mapping generally does not need a `url`
 - The replacement should have a `url` to its external documentation (e.g. MDN or Node docs)
-- If it's a web standard, then the replacement should have a `webFeatureId` (these come from the [`web-features` dataset](https://github.com/web-platform-dx/web-features/tree/main/features))
-- If it's a Node built-in, then the replacement should have a `nodeFeatureId` (these are of the form `{ moduleName: string, exportName?: string }`. For example, `fs` would be `{"moduleName": "fs"}`, while `fs.readFile` would be `{"moduleName": "fs", "exportName": "readFile"}`)
+- If it's a web standard, then the replacement should have a `webFeatureId` (these come from the [`web-features` dataset](https://github.com/web-platform-dx/web-features/tree/main/features)), and a `url` to MDN
+- If it's a Node built-in, then the replacement should have a `nodeFeatureId` (these are of the form `{ moduleName: string, exportName?: string }`. For example, `fs` would be `{"moduleName": "fs"}`, while `fs.readFile` would be `{"moduleName": "fs", "exportName": "readFile"}`), and a `url` to the Node docs
 
 When using the `simple` type:
 


CONTRIBUTING.md
```diff
@@ -51,20 +51,20 @@ There are different types of replacements available:
 - `removal`: for removing the old module without a replacement (e.g. modules which check if a now widely supported API is available)
 - `simple`: for replacing the old module with inline code
 - `native`: for replacing the old module with native APIs (e.g. standard web APIs, or Node built-ins)
-- `documented`: for replacing the old module with a documented alternative
+- `documented`: a guide for replacing the old module with a new module and/or code
```
gameroman: I think this follow the other ones that start with `for replacing`

CONTRIBUTING.md
```diff
@@ -51,20 +51,20 @@ There are different types of replacements available:
...
 
 When using the `native` type:
 
-- The mapping generally does not need a `url`
+- The mapping generally does not need a `url`, but if it does then it's because ...
 - The replacement should have a `url` to its external documentation (e.g. MDN or Node docs)
-- The replacement should have a `webFeatureId` if it is a web standard (these come from the [`web-features` dataset](https://github.com/web-platform-dx/web-features/tree/main/features))
-- The replacement should have a `nodeFeatureId` if it is a Node built-in (these are of the form `{ moduleName: string, exportName?: string }`. For example, `fs` would be `{"moduleName": "fs"}`, while `fs.readFile` would be `{"moduleName": "fs", "exportName": "readFile"}`)
+- If it's a web standard, then the replacement should have a `webFeatureId` (these come from the [`web-features` dataset](https://github.com/web-platform-dx/web-features/tree/main/features))
+- If it's a Node built-in, then the replacement should have a `nodeFeatureId` (these are of the form `{ moduleName: string, exportName?: string }`. For example, `fs` would be `{"moduleName": "fs"}`, while `fs.readFile` would be `{"moduleName": "fs", "exportName": "readFile"}`)
```
gameroman: I'm thinking something like this

```suggestion
- If it's a web standard, then the replacement should have a `webFeatureId` (these come from the [`web-features` dataset](https://github.com/web-platform-dx/web-features/tree/main/features)) and  `url` to MDN
- If it's a Node built-in, then the replacement should have a `nodeFeatureId` (these are of the form `{ moduleName: string, exportName?: string }`. For example, `fs` would be `{"moduleName": "fs"}`, while `fs.readFile` would be `{"moduleName": "fs", "exportName": "readFile"}`) and `url` to Node docs
```

CONTRIBUTING.md
ghostdevv: I tried but I couldn't get the words right :melting_face: any ideas?

ghostdevv: @43081j said I'm fired because of this horrendous incorrect suggestion so I've removed it /j

CONTRIBUTING.md
ghostdevv: So apparently there's not a mapping that has a `url` and so @43081j worded it that way to leave the door open but not knowing what there would be. The MDN/Node docs link would go on the replacement itself

ghostdevv: oh this is about the replacement lol

ghostdevv: done

gameroman (approved):
lgtm

PR #481, opened 2026-03-26
fix: update types

gameroman:
### 📚 Description

Add `description?: never` to `DocumentedModuleReplacement` for easier use with typescript

This is to avoid doing things like this:

https://github.com/npmx-dev/npmx.dev/blob/d80c709c30b41ac68bb91b6524fabe59477417d2/app/components/Package/Replacement.vue#L17-L19

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -21,6 +21,7 @@ interface ModuleReplacementLike {
 export interface DocumentedModuleReplacement extends ModuleReplacementLike {
   type: 'documented';
   url: KnownUrl;
+  description?: never;
   replacementModule: string;
 }
 


gameroman:
In that case we can add `description?: string` to `ModuleReplacementLike` instead

43081j:
we're not comparing `never` and `undefined` though.

```ts
// this
interface X {
  prop?: never;
}

// is not the same as this
interface X {
  // no prop at all
}
```

the types are correct right now and are doing their job by making you narrow the union correctly.

it may feel like a pain but the documented replacements don't have a description property at all. setting them to `never` is just a workaround to avoid having to narrow the type.

gameroman:
But why would you want to narrow the type for the description that is already optional on other types?

gameroman:
What would be yours proposed solution?

PR #488, opened 2026-04-02
feat: add `getReplacementDescription` util

gameroman:
### 📚 Description

Add `getReplacementDescription` util

See #481 and [npmx-dev/npmx.dev#2068 (comment)](https://github.com/npmx-dev/npmx.dev/pull/2068/#discussion_r3012590285)

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -1,4 +1,4 @@
-import type {KnownUrl} from './types.js';
+import type {KnownUrl, ModuleReplacement} from './types.js';
 
 export function resolveDocUrl(url?: KnownUrl): string | null {
   if (!url) return null;
@@ -14,4 +14,11 @@ export function resolveDocUrl(url?: KnownUrl): string | null {
     default:
       return null;
   }
+}
+
+export function getReplacementDescription(
+  replacement: ModuleReplacement
+): string | undefined {
+  if (!('description' in replacement)) return undefined;
+  return replacement.description;
 }


src/util.ts
```diff
@@ -15,3 +15,10 @@ export function resolveDocUrl(url?: KnownUrl): string | null {
       return null;
   }
 }
+
+export function getReplacementDescription(
+  replacement: ModuleReplacement
+): string | undefined {
```
43081j, line 21:
```suggestion
  switch (replacement.type) {
    case 'documented':
      return undefined;
    default:
      return replacement.description;
  }
```

we should narrow it rather than checking the shape manually i think

gameroman: I'm thinking this is better because if we ever add `description` to `documented` type we won't need to change this function and don't accidentally forget to

43081j: i think we should have to update the code.

its better to narrow the union correctly instead of narrowing the shape manually.

this description stuff seems to be eating up more time than it should be. its specifically a tagged union so we are made to narrow the type. the correct way to narrow the type is this switch. all of this came from you wanting to save some lines of code (wanting all replacements to have a description so you don't have to do any narrowing). but that would've meant pretending documented replacements have a description just to save some bytes, which isn't right. this way, if we ever do introduce a description, it isn't a breaking change.

gameroman: Added the util to npmx pr instead 👍

PR #504, opened 2026-04-03
fix: check that all mappings are used

gameroman:
### 🔗 Linked issue

See https://github.com/es-tooling/module-replacements/pull/312#issuecomment-4185494733

### 📚 Description

- Added a check that all mappings are used
- Fixed associated finding
- Split formatting and module validation in ci for easier overview

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,7 +23,5 @@ jobs:
         run: npm ci
       - name: Build
         run: npm run build
-      - name: Check formatting
-        run: npm run lint:format
-      - name: Validate manifests
-        run: npm run lint:manifests
+      - name: Lint
+        run: npm run lint


ghostdevv (dismissed):

.github/workflows/main.yml
```diff
@@ -23,5 +23,7 @@ jobs:
         run: npm ci
       - name: Build
         run: npm run build
-      - name: Lint
-        run: npm run lint
```
43081j, line 27: can you explain this one?

`lint` already runs both. im not sure we need individual steps

gameroman: It just makes it easier to see which part failed in overview

43081j: the risk is that we add a third lint script in future and have to remember to add it here. having it in one means we don't need to remember that.

thoughts?

gameroman: > the risk is that we add a third lint script in future and have to remember to add it here. having it in one means we don't need to remember that.
> 
> thoughts?

True, but might be it is a good idea to have those decoupled

We already do that in npmx, for example

43081j: might it? can you explain why?

by having `npm run lint` in here, they are decoupled since we only need to add new lint scripts to the main `lint` script.

i'm pretty sure we don't need to make this change and can just look at the result if we want to know what failed.

PR #517, opened 2026-04-07
feat: add `as-array` to replacements

### LINKED ISSUE #515: [Replacement]: `as-array`
### Package to replace

`as-array`

### Suggested replacemen

```js
function asArray(data) {
  if (data == null) return []
  if (Array.isArray(data)) return data
  if (typeof data !== 'string' && data[Symbol.iterator]) return Array.from(data)
  return [data]
}
```

### Manifest type

### Rationale

Should not be its own package

### Availability

all modern browsers and runtimes

gameroman:
### 🔗 Linked issue

Closes #515

### 📚 Description

add `as-array` to replacements

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -4,7 +4,7 @@
       "id": "snippet::array-coerce",
       "type": "simple",
       "description": "You can use a combination of a ternary operator and `Array.isArray` to make sure a value, `undefined`, `null` or an array is always returned as an array.",
-      "example": "(val == null ? [] : Array.isArray(val) ? val : [val])\n// Or if you need to convert iterable into array\nArray.from(iterable)"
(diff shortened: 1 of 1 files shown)


manifests/micro-utilities.json
```diff
@@ -4,7 +4,7 @@
       "id": "snippet::array-coerce",
       "type": "simple",
       "description": "You can use a combination of a ternary operator and `Array.isArray` to make sure a value, `undefined`, `null` or an array is always returned as an array.",
-      "example": "(arr == null ? [] : Array.isArray(arr) ? arr : [arr])"
+      "example": "(val == null ? [] : Array.isArray(val) ? val : [val])\n// Or if you need to convert iterable into array\nArray.from(iterable)"
```
```suggestion
      "example": "(val == null ? [] : Array.isArray(val) ? val : [val])\n// Or if you need to convert an iterable into an array\nArray.from(iterable)"
```

PR #536, opened 2026-04-08
feat: add `fs-then-native` to native manifest

### LINKED ISSUE #520: [Replacement]: `fs-then-native`
### Package to replace

fs-then-native

### Suggested replacement(s)

`node:fs/promises`

### Manifest type

native (replaceable by a built-in platform feature)

### Rationale

The package is a pseudo-polyfill for a feature now widely available

### Availability

Node v14.0.0+ (the api existed a few versions before, but not under the same export)

### Code example (optional)

```js

```

gameroman:
### 🔗 Linked issue

Closes #520

### 📚 Description

add `fs-then-native` to native manifest

### DIFF
--- a/manifests/native.json
+++ b/manifests/native.json
@@ -1489,6 +1489,12 @@
       },
       "nodeFeatureId": {"moduleName": "fs"}
     },
+    "fsPromises": {
+      "id": "fsPromises",
+      "type": "native",
+      "url": {"type": "node", "id": "api/fs.html#promises-api"},
+      "nodeFeatureId": {"moduleName": "fs/promises"}
+    },
     "globalThis": {
       "id": "globalThis",
       "type": "native",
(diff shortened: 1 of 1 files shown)


manifests/native.json
```diff
@@ -1489,6 +1489,12 @@
       },
       "nodeFeatureId": {"moduleName": "fs"}
     },
+    "fsPromises": {
+      "id": "fsPromises",
+      "type": "native",
+      "url": {"type": "node", "id": "api/fs.html#promises-api"},
+      "nodeFeatureId": {"moduleName": "fs/promises"}
```
43081j: this isn't right since async FS doesn't have its own documented module - it is part of `fs`.

the right `moduleName` is `fs`

gameroman: Can this be validated automatically somehow?

43081j: not really since we can't possibly know what doc you meant to link to.

we could check if it 404s but that means we'd need to `fetch` each URL which isn't great (as part of `lint`).

for now maybe just move on as this has only happened once and is what code review is for

43081j (approved):

PR #540, opened 2026-04-08
fix: check that a replacement is not mapped to itself

### LINKED ISSUE #526: [Replacement]: `dottie`
### Package to replace

[`dottie`](https://npmx.dev/package/dottie)

### Suggested replacement(s)

[`dlv`](https://e18e.dev/docs/replacements/dot-prop)

### Manifest type

### Rationale

`dottie` has been deprecated

### Availability

_No response_

### Code example (optional)

```js

```

### LINKED ISSUE #534: `object-path` is listed as a replacement for itself
`object-path` has suggested replacements, one of the replacements is `object-path` itself

gameroman:
### 🔗 Linked issue

Closes #526
Closes #534

### 📚 Description

check that a replacement is not mapped to itself

### DIFF
--- a/docs/modules/dot-prop.md
+++ b/docs/modules/dot-prop.md
@@ -20,7 +20,7 @@ setProperty(obj, 'foo.bar.baz', 'value') // [!code --]
 dset(obj, 'foo.bar.baz', 'value') // [!code ++]
 ```
 
-## `.reduce` (native)
+## `String.prototype.split` + `Array.prototype.reduce` (native)
 
 If you only need to get a nested value you can use a simple function:
 
@@ -31,3 +31,7 @@ function getProperty(obj, key) {
 
 const value = getProperty(obj, 'foo.bar.baz')
 ```
+
+> [!NOTE]
+> This assumes that you do not consume dot paths as user input.
+> If you do, ensure you sanitize keys before accessing them (e.g. through `Object.hasOwn`).
--- a/docs/modules/dot-prop.md
+++ b/docs/modules/dot-prop.md
@@ -20,23 +20,14 @@ setProperty(obj, 'foo.bar.baz', 'value') // [!code --]
 dset(obj, 'foo.bar.baz', 'value') // [!code ++]
 ```
 
-## `object-path`
+## `.reduce` (native)
 
-[`object-path`](https://github.com/mariocasciaro/object-path) provides get/set/has/delete operations plus array methods like push, insert, and empty.
+If you only need to get a nested value you can use a simple function:
 
-```ts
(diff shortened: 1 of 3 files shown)


docs/modules/dot-prop.md
```diff
@@ -20,23 +20,14 @@ setProperty(obj, 'foo.bar.baz', 'value') // [!code --]
 dset(obj, 'foo.bar.baz', 'value') // [!code ++]
 ```
 
-## `object-path`
+## `.reduce` (native)
```
43081j:
```suggestion
## `Array.prototype.reduce` (native)
```

docs/modules/dot-prop.md
```diff
@@ -20,23 +20,14 @@ setProperty(obj, 'foo.bar.baz', 'value') // [!code --]
...
-const exists = objectPath.has(obj, 'foo.bar.baz') // [!code ++]
+```js
+function getProperty(obj, key) {
+  return key.split('.').reduce((acc, k) => acc?.[k], obj)
+}
 
-deleteProperty(obj, 'foo.bar.baz') // [!code --]
-objectPath.del(obj, 'foo.bar.baz') // [!code ++]
+const value = getProperty(obj, 'foo.bar.baz')
 ```
```
43081j, line 33:
````suggestion
```

Note that this is assuming you do not consume dot paths as user input. If you do, ensure you sanitize keys before accessing them (e.g. through `Object.hasOwn`).
````

File: manifests/preferred.json [hunk 3]
```diff
@@ -183,7 +183,7 @@
     "dot-prop": {
       "type": "module",
       "moduleName": "dot-prop",
-      "replacements": ["dlv", "object-path"],
+      "replacements": ["dlv", "dset"],
```
43081j: this isn't quite right.

dlv and dset do different jobs, but by having both in this array, you're saying they are equal replacements to `dot-prop`.

probably best to leave it as `dlv` only and mention `dset` in the doc

gameroman: I guess it doesn't matter since they link to the same doc anyway?

43081j: it does matter. we only show one replacement in messages (cli, npmx, etc). the first one that matches

but regardless, this logically doesn't make sense since it implies `dlv` and `dset` do the same job, and are equal replacements to `dot-prop`, which isn't true.

gameroman: But then `dlv` is also not a full replacement on its own

43081j: yup that is correct.

we don't have a way to represent "partial replacements". so best effort is to list `dlv` and explain the rest in the doc.

we can discuss/figure out partial replacements separate from this

43081j (approved):

PR #591, opened 2026-04-13
chore: make documented URLs optional

43081j:
### 🔗 Linked issue

N/A

### 📚 Description

This makes documented replacements have an optional URL rather than it
being required. Most or all documented replacements are part of a
mapping which itself has the unified documentation (of all replacements
for that mapping). So we don't really need to repeat the same URL per
replacement.

### DIFF
--- a/manifest-schema.json
+++ b/manifest-schema.json
@@ -124,8 +124,7 @@
       "required": [
         "id",
         "replacementModule",
-        "type",
-        "url"
+        "type"
       ],
       "additionalProperties": false
     },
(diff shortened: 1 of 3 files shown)


43081j:
@ghostdevv @gameroman thoughts on this?

ghostdevv (approved):
Sounds good - we should really make some linting rules for the replacements, one of which could make sure that either the replacement or mapping has docs when relevant

gameroman (approved):
This makes sense 👍

43081j:
> Sounds good - we should really make some linting rules for the replacements, one of which could make sure that either the replacement or mapping has docs when relevant

this is a good idea 👍 

we do already have some rules, and roman added a few recently. so would be good to add this one too

Issue #553, opened 2026-04-09
Possible adjustments for renovate-config


OrbisK:
I think we need to make some minor changes to improve a possible renovate-presets  (proof of concept https://github.com/OrbisK/renovate-config-e18e).

## `https://e18e.dev/docs/replacements/<pgk>`

We should consider using the same doc path as the module name. Or, at the very least, we should implement redirects to cover the original name. This is because the renovate presets have templating that can use the `depName`. Having to map it to a doc path means writing a package rule for each doc path.

```
      prBodyNotes: [
        `{{#unless (includes prBodyColumns "Community Notes")}}> [!WARNING]
> **This package has a recommended replacement.** Check the [e18e replacement guide for \`{{{depName}}}\`](https://e18e.dev/docs/replacements/{{{depName}}}) to find modern, lighter alternatives.{{/unless}}`,
      ],
```

## Each replacement should have one preferred/optionated default replacement.

In order for Renovate to open a replacement PR, one package is needed to replace it. (We need to make some changes to package.json to open a replacement PR.)

## Publish the docs md

We should publish the MD docs to embed them in Renovate if necessary.

To make this work, we must remove all VitePress-only markdown flavour (e.g. `//--code? ) and use normal markdown diffs.

OrbisK:
> I think we cant check the (node ?) engine in renvoate

there is `matchJsonata`, but this cant compare semver i guess. we can only tell if someone has exact strings as node engine + it  can only check package.json engines not the node version of pipelines, dockerfiles etc

43081j:
You need the pre release (`@next`). Each `mappings` entry has an array of `replacements`. Each is a key of one of the objects in the root `replacements` object.

OrbisK:
> You need the pre release (`@next`). Each `mappings` entry has an array of `replacements`. Each is a key of one of the objects in the root `replacements` object.

Done https://github.com/OrbisK/renovate-config-e18e/commit/b581163763867a6ba8bc0edb4fdb3134425d682a#diff-fefbbfc7ff66fae66581bec368540ca31b7f5dfca81e2ae4c68248fb6ab3afa1

Now I "only " need the docs and it should be mostly done 😅

gameroman:
Should this be closed now?

43081j:
let's wait and see what @OrbisK says. i think hes been busy with other things lately. would love to catch up on the renovate integration though

OrbisK:
[...]

PR #592, opened 2026-04-13
docs: replace all vitepress flavored markdown

OrbisK:
### 🔗 Linked issue

related #553 

### 📚 Description

Replacing vitepress flavor makes the markdown github/gitlab compartible. (this removes lang highlighting for diffs 😞) 

Disclaimer: 90% of this replacement is done by AI.

### DIFF
--- a/docs/modules/body-parser.md
+++ b/docs/modules/body-parser.md
@@ -10,18 +10,18 @@ description: Modern alternatives to the body-parser package for parsing HTTP req
 
 Example:
 
-```ts
-import bodyParser from 'body-parser' // [!code --]
-import { json, urlencoded } from 'milliparsec' // [!code ++]
-import express from 'express'
+```diff
+- import bodyParser from 'body-parser'
++ import { json, urlencoded } from 'milliparsec'
+  import express from 'express'
 
-const app = express()
+  const app = express()
 
-app.use(bodyParser.json()) // [!code --]
-app.use(bodyParser.urlencoded({ extended: true })) // [!code --]
+- app.use(bodyParser.json())
+- app.use(bodyParser.urlencoded({ extended: true }))
 
-app.use(json()) // [!code ++]
-app.use(urlencoded()) // [!code ++]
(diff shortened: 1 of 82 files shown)


43081j:
does vitepress support `diff` as a language?

i.e. is it exactly equivalent to vitepress' `code ++` etc?

gameroman:
I’m not sure we should do this

OrbisK:
> i.e. is it exactly equivalent to vitepress' `code ++` etc?

we lose the lang highlighting. diff is not lang specific.

OrbisK:
> I’m not sure we should do this

Alternatively I can strip it with some regex magic 😅

OrbisK:
plain diff looks like this: 
image

gameroman:
Maybe it could be simpler to link to the doc instead

ghostdevv:
I'd also like to avoid loosing the syntax highlighting if possible - [expressive code can do both](https://expressive-code.com/key-features/text-markers/#combining-syntax-highlighting-with-diff-like-syntax) so I wonder if we could patch that in :thinking:

OrbisK:
> Maybe it could be simpler to link to the doc instead

Mhh. I think it feels more renovate native to show it inline. like changelog etc.


> I'd also like to avoid loosing the syntax highlighting if possible - [expressive code can do both](https://expressive-code.com/key-features/text-markers/#combining-syntax-highlighting-with-diff-like-syntax) so I wonder if we could patch that in 🤔

could work.

gameroman:
> does vitepress support `diff` as a language?

Doesn't seem so

OrbisK:
> plain diff looks like this: image

This is from my vitepress project 🤔

43081j:
i think you're going to have to do this in a transform when consuming the docs.

see this screenshot where i changed one and not the other:

image

see the vitepress diff is much richer than the diff lang, since the latter is just a flavour of highlighting rather than an actual component in the end.

PR #595, opened 2026-04-14
feat: add `base64id` to replacements

gameroman:
### 🔗 Linked issue

Closes #552

### 📚 Description

add `base64id` to replacements

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -66,6 +66,12 @@
       "description": "Every modern runtime provides a way to convert byte array to and from base64.",
       "example": "// From base64 to Uint8Array\nconst bytes = Uint8Array.fromBase64(base64)\n// From Uint8Array to base64\nconst base64 = bytes.toBase64()"
     },
+    "snippet::base64-id": {
+      "id": "snippet::base64-id",
+      "type": "simple",
+      "description": "You can use `crypto.randomBytes` with `Buffer.prototype.toString` to generate a random base64 id",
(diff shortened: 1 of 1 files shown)


43081j (approved):

Dustin4444 (commented):
@ #600

manifests/micro-utilities.json
```diff
@@ -66,6 +66,12 @@
       "description": "Every modern runtime provides a way to convert byte array to and from base64.",
       "example": "// From base64 to Uint8Array\nconst bytes = Uint8Array.fromBase64(base64)\n// From Uint8Array to base64\nconst base64 = bytes.toBase64()"
     },
+    "snippet::base64-id": {
+      "id": "snippet::base64-id",
+      "type": "simple",
+      "description": "You can use `crypto.randomBytes` with `Buffer.prototype.toString` to generate a random base64 id",
+      "example": "import crypto from 'node:crypto'\nconst id = crypto.randomBytes(15).toString('base64').replaceAll('+', '-').replaceAll('/', '_')"
```
SuperchupuDev, line 73: sorry for the late comment but we should probably recommend `Uint8Array` over `Buffer` due to it being a built-in javascript feature, now that it has base64 support

sindre has a blog post that goes in depth about this topic: https://sindresorhus.com/blog/goodbye-nodejs-buffer

example with `Uint8Array` + web crypto api (no imports needed):
```js
crypto.getRandomValues(new Uint8Array(15)).toBase64();
```

gameroman: Since the original package also uses `node:crypto` module I decided to keep it simple

43081j: can one of you open an issue so we can look into it?

i think you may be right and we should just use uint8arrays but lets track it in a new issue

PR #639, opened 2026-04-28
feat(preferred): add `sqlite3` replacement options

### LINKED ISSUE #623: [Replacement]: `sqlite3`
### Package to replace

sqlite3

### Suggested replacement(s)

sqlite

### Manifest type

### Rationale

The package is unmaintained and the suggested alternative is a maintained fork
Seems to have good compatibility so long as promises are used rather than callbacks

### Availability

_No response_

### Code example (optional)

```js

```

dreyfus92:
### 📚 Description

this PR adds `sqlite3` replacements to the preferred manifest in the agreed order (`node:sqlite`, `better-sqlite3`, `sqlite`), includes the matching replacement metadata entries, and adds a new docs page with quick migration examples.

closes: #623

### DIFF
--- a/docs/modules/sqlite3.md
+++ b/docs/modules/sqlite3.md
@@ -1,0 +1,56 @@
+---
+description: Modern alternatives to the sqlite3 package for working with SQLite in Node.js
+---
+
+# Replacements for `sqlite3`
+
+## `node:sqlite` (native, Node.js)
+
+Node.js ships a built-in SQLite module, [`node:sqlite`](https://nodejs.org/api/sqlite.html), which is the preferred option when you can target a recent Node runtime.
+
+Example:
+
+```ts
+import sqlite3 from 'sqlite3' // [!code --]
+import { DatabaseSync } from 'node:sqlite' // [!code ++]
+
+const db = new sqlite3.Database('app.db') // [!code --]
+const db = new DatabaseSync('app.db') // [!code ++]
+
+db.all('SELECT * FROM users', (err, rows) => {}) // [!code --]
+const rows = db.prepare('SELECT * FROM users').all() // [!code ++]
+```
+
+## `better-sqlite3`
+
+[`better-sqlite3`](https://github.com/WiseLibs/better-sqlite3) is a popular and actively maintained SQLite library for Node.js.
+
+Example:
+
+```ts
+import sqlite3 from 'sqlite3' // [!code --]
+import Database from 'better-sqlite3' // [!code ++]
+
+const db = new sqlite3.Database('app.db') // [!code --]
+const db = new Database('app.db') // [!code ++]
+
+db.all('SELECT * FROM users', (err, rows) => {}) // [!code --]
+const rows = db.prepare('SELECT * FROM users').all() // [!code ++]
+```
+
+## `sqlite`
+
+[`sqlite`](https://github.com/kriasoft/node-sqlite) is an actively maintained option commonly used for promise-based APIs.
+
+Example:
+
+```ts
+import sqlite3 from 'sqlite3' // [!code --]
+import { open } from 'sqlite' // [!code ++]
+
+const db = new sqlite3.Database('app.db') // [!code --]
+const db = await open({ filename: 'app.db', driver: sqlite3.Database }) // [!code ++]
+
+db.all('SELECT * FROM users', (err, rows) => {}) // [!code --]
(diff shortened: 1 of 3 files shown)


docs/modules/sqlite3.md
```diff
@@ -0,0 +1,56 @@
...
+```ts
+import sqlite3 from 'sqlite3' // [!code --]
+import { open } from 'sqlite' // [!code ++]
+
+const db = new sqlite3.Database('app.db') // [!code --]
+const db = await open({ filename: 'app.db', driver: sqlite3.Database }) // [!code ++]
+
+db.all('SELECT * FROM users', (err, rows) => {}) // [!code --]
+const rows = await db.all('SELECT * FROM users') // [!code ++]
+```
```
gameroman: This doesn't seem to be a replacement if it still requires `sqlite3` dependency. Should we remove it from recommendations?

43081j: seems you are right. `sqlite` is a client only, and `sqlite3` is a driver. good, confusing naming.

ghostdevv:
`sqlite` doesn't appear to be used much (relative to better-sqlite3), and is using a one major version behind version of `sqlite3` - is it a suitable replacement? actually a shame it's not used much since, on paper it otherwise seems more modern

<a href="https://npmx.dev/compare?packages=sqlite,better-sqlite3">

</a>

43081j:
note this @ghostdevv 
https://github.com/e18e/module-replacements/pull/639#discussion_r3166528352

sqlite is actually just a client but still needs a driver: sqlite3.

so we should remove it from the list here

EDIT: i removed it and caught up from main

manifests/micro-utilities.json
```diff
@@ -100,7 +100,7 @@
       "id": "snippet::has-ansi",
       "type": "simple",
       "description": "You can use the `includes` method on the string to check if a specific ANSI byte is present.",
-      "example": "string.includes(\"\u001B\") || string.includes(\"\u009B\")"
+      "example": "string.includes(\"\u001b\") || string.includes(\"\")"
```
```suggestion
      "example": "string.includes(\"\\u001B\") || string.includes(\"\\u009B\")"
```

docs/modules/sqlite3.md
```diff
@@ -0,0 +1,39 @@
+---
+description: Modern alternatives to the sqlite3 package for working with SQLite in Node.js
+---
+
+# Replacements for `sqlite3`
+
+## `node:sqlite` (native, Node.js)
```
gameroman:
```suggestion
## `node:sqlite` (native, since Node.js 22.13.0)
```

gameroman (changes requested):

manifests/micro-utilities.json
ghostdevv: should this be done in a separate PR and undo the change here?

docs/modules/sqlite3.md
ghostdevv: In `22.5.0` it's under the flag `--experimental-sqlite`, they unflag it in `22.13.0`

https://nodejs.org/docs/latest-v22.x/api/sqlite.html#sqlite

manifests/micro-utilities.json
43081j: the formatter is doing it FYI, it isn't a change by me or paul

it seems prettier prints a x9B literally instead of an escape sequence

ghostdevv: but why is the formatting changing on this PR? Lint is running on main? :thinking:

43081j: the sort-manifests script runs prettier. in CI we also run `prettier --check` to fail CI if it isn't formatted

this is just the result of running the formatter. prettier being a bit weird since it didn't do it before

ghostdevv: Just wanna understand what is causing the formatting change when `pretter --check` is passing everywhere else :thinking:

gameroman: It was originally, it should be a double `\\` to escape the `\` and show `string.includes("\u001B") || string.includes("\u009B")` to the consumer

docs/modules/sqlite3.md
gameroman: Updated 👍

manifests/micro-utilities.json
43081j: ah yeah you are right. the example was wrong before this branch

43081j (approved):

PR #641, opened 2026-04-29
feat: update Node.js version matrix to include lts/*

### LINKED ISSUE #619: Test Node v24 as well as latest
Right now, CI hits Node v20 and v22, and then latest (which is 25). It should probably also hit v24? https://github.com/e18e/module-replacements/blob/main/.github/workflows/main.yml

dreyfus92:
### 📚 Description
tweaks the main CI to track LTS

closes: #619

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [20.x, 22.x, latest]
+        node-version: [20.x, 22.x, lts/*, latest]
       fail-fast: false
     steps:
       - name: Checkout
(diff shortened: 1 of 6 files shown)


.github/workflows/main.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [20.x, 22.x, latest]
+        node-version: [20.x, 22.x, lts/*, latest]
```
43081j, line 13: presumably lts currently includes 20?

we don't really want to remove the explicit `20.x` though as that'd mean our tested versions is a moving window.

it may be better to specify 25 directly 🤔

manifests/native.json
```diff
@@ -2217,6 +2217,11 @@
       "moduleName": "indexof",
       "replacements": ["Array.prototype.indexOf"]
     },
+    "is-arrayish": {
```
43081j, line 2247: what happened here? 👀

dreyfus92:
gonna open a new pr 😂

PR #650, opened 2026-05-05
refactor: switch to pnpm

ghostdevv:
### 🔗 Linked issue

n/a

### 📚 Description

Switch the repo to use pnpm as it's currently npm. I picked pnpm 10 as we still support Node 20 and pnpm 11 doesn't.


joaopedrodcf (approved):
Maybe we can in another PR increase the version of node to 22 and migrate to pnpm 11

ghostdevv:
> Maybe we can in another PR increase the version of node to 22 and migrate to pnpm 11

We could definitely consider bumping v3 to 22 now 20 is EOL! Let's think about it in a follow up as we need to consider consumers of `module-replacements`, but I think it should be ok since v3 beta already bumped from 18 to 20

gameroman (approved):

dreyfus92 (approved):
🚢

ghostdevv:
yolo

PR #686, opened 2026-05-29
fix: use fs/promises for fs.rm

Netail:
### 🔗 Linked issue

No issue, but https://discord.com/channels/1227627367204257852/1227653370945474590/1510035958907666685

### 📚 Description

Align with https://github.com/e18e/module-replacements/blob/main/manifests/preferred.json#L3065-L3070

### DIFF
--- a/manifests/native.json
+++ b/manifests/native.json
@@ -1579,8 +1579,8 @@
     "fs.rm": {
       "id": "fs.rm",
       "type": "native",
-      "url": {"type": "node", "id": "api/fs.html#fsrmpath-options-callback"},
-      "nodeFeatureId": {"moduleName": "fs", "exportName": "rm"}
+      "url": {"type": "node", "id": "api/fs.html#fspromisesrmpath-options"},
+      "nodeFeatureId": {"moduleName": "node:fs/promises", "exportName": "rm"}
     },
     "fsPromises": {
       "id": "fsPromises",


manifests/native.json
```diff
@@ -1579,8 +1579,8 @@
     "fs.rm": {
       "id": "fs.rm",
```
ghostdevv, line 1580: we should probably also change this to `fs.promises.rm` for consistency - otherwise LGTM

ghostdevv (changes requested):

manifests/native.json
Netail: Same for the other replacement entry in the preferred list?

gameroman (approved):

manifests/native.json
ghostdevv: Yea if it's not too much :smile:

43081j: worth noting no other entry has `.promises` in its ID yet, even the ones which are from `node:fs/promises`.

but technically you're probably right in case we ever had `fs.rm` (sync) in here for whatever reason

gameroman: Also if we are going to do this, probably makes sense changing `fsPromises` to `fs.promises`

ghostdevv: shit sorry Roman I accidentally deleted your comment while trying to negotiate with GitHub - what I was trying to say was let's do this in a follow up :smile:

ghostdevv (approved):

manifests/native.json
ghostdevv: https://github.com/e18e/module-replacements/pull/687

PR #720, opened 2026-06-08
fix: first node version that supports `fs.mkdtempDisposable`

Kiwow:
### 📚 Description

According to [the Node docs](https://nodejs.org/api/fs.html#fspromisesmkdtempdisposableprefix-options), `mkdtempDisposable` was added in Node 24.4.0, not 20.4.0 as we say currently.

This also tracks with the fact that Node only added support for the explicit resource managenment and the `using` keyword in version 24.


gameroman (approved):
Looks like that was my typo 😄

dreyfus92 (approved):
nice catch @Kiwow 😄, thank you

Kiwow:
Happy to help, thanks for the quick turnaround and for your guys' work on this project!

PR #722, opened 2026-06-09
feat(preferred): add tsdown as documented replacement

### LINKED ISSUE #649: Add replacement: `tsup` → `tsdown`
Following up from the discussion in [e18e/ecosystem-issues#170](https://github.com/e18e/ecosystem-issues/issues/170), I'd like to propose adding `tsup` → `tsdown` to the module-replacements dataset.

### Basic info

- **Module to replace:** [`tsup`](https://www.npmjs.com/package/tsup)
- **Replacement:** [`tsdown`](https://www.npmjs.com/package/tsdown)
- **Manifest:** `preferred`
- **Replacement type:** `documented`

### Why it should be replaced

- **Upstream recommends it.** tsup's own README has [recommended migrating to tsdown since August 2025](https://github.com/egoist/tsup/commit/8b6907df5d944e5e4a639a297875a35818ba9fa5). The project is effectively in maintenance mode.
- **Performance.** tsdown is built on Rolldown (Rust-based) and is significantly faster than tsup for typical library builds, while producing comparable or smaller output.
- **Size.** tsdown's install footprint is now comparable to tsup (~22–26MB vs ~25–35MB) and shrinking further as the dts plugin is integrated natively.
- **Significant adoption.** tsup has tens of millions of weekly downloads. tsdown already powers builds in major projects including Vite, Slidev, RSSHub, Rolldown itself, unplugin, and 600+ other projects with >100 stars.
- **Drop-in compatible.** tsdown is designed as a tsup successor with a compatible config surface; most projects migrate with little to no config change.

### Availability

- tsdown supports Node.js ≥ 20.19 (current active LTS line), aligning with modern tooling expectations.
- Available on npm, no native/system dependencies beyond what Rolldown ships.

### Migration example

Most `tsup.config.ts` files map 1:1 to `tsdown.config.ts`:

```ts
// tsup.config.ts
import { defineConfig } from 'tsup'

export default defineConfig({
  entry: ['src/index.ts'],
  format: ['esm', 'cjs'],
  dts: true,
  clean: true,
})
```

```ts
// tsdown.config.ts
import { defineConfig } from 'tsdown'

export default defineConfig({
  entry: ['src/index.ts'],
  format: ['esm', 'cjs'],
  dts: true,
  clean: true,
})
```

Tooling is also available to automate this:

- **CLI:** `npx tsdown-migrate`
- **AI skill:** `npx skills add rolldown/tsdown --skill tsdown-migrate`

### Scope

To echo the concern raised by @43081j in the linked discussion: this replacement is intended only for **existing tsup users**. It should not be read as a recommendation for projects on other bundlers, nor for projects that don't need a bundler at all.

Happy to open the PR once the direction here is agreed upon.

dreyfus92:
### 🔗 Linked issue

closes: #649

### 📚 Description

 adds `tsup` → `tsdown` as a `documented` replacement in the `preferred` manifest and migration docs with scope note

### DIFF
--- a/docs/modules/tsup.md
+++ b/docs/modules/tsup.md
@@ -1,0 +1,49 @@
+---
+description: Modern alternative to tsup for bundling TypeScript libraries
+---
+
+# Replacements for `tsup`
+
+## Scope
+
+This replacement is for **existing `tsup` users** only. It is not a recommendation for projects on other bundlers, or for projects that do not need a bundler.
+
+## `tsdown`
+
+[tsdown](https://github.com/rolldown/tsdown) is the successor recommended by [tsup's own README](https://github.com/egoist/tsup) since August 2025. The project is effectively in maintenance mode. tsdown is built on Rolldown and is typically faster for library builds, with a largely drop-in config surface.
+
+### Config migration
+
+Most `tsup.config.ts` files map 1:1 to `tsdown.config.ts`:
+
+```ts
+import { defineConfig } from 'tsup' // [!code --]
+import { defineConfig } from 'tsdown' // [!code ++]
+
+export default defineConfig({
+  entry: ['src/index.ts'],
+  format: ['esm', 'cjs'],
+  dts: true,
+  clean: true
+})
+```
+
+### Package scripts
+
+```json
+{
+  "scripts": {
+    "build": "tsup" // [!code --]
+    "build": "tsdown" // [!code ++]
+  }
+}
+```
+
+### Automated migration
+
+- `npx tsdown-migrate`
+- `npx skills add rolldown/tsdown --skill tsdown-migrate`
+
+### Requirements
+
+- Node.js >= 20.19 (current active LTS line)
--- a/docs/modules/tsup.md
+++ b/docs/modules/tsup.md
@@ -4,10 +4,6 @@ description: Modern alternative to tsup for bundling TypeScript libraries
 
 # Replacements for `tsup`
 
-## Scope
-
-This replacement is for **existing `tsup` users** only. It is not a recommendation for projects on other bundlers, or for projects that do not need a bundler.
-
 ## `tsdown`
 
 [tsdown](https://github.com/rolldown/tsdown) is the successor recommended by [tsup's own README](https://github.com/egoist/tsup) since August 2025. The project is effectively in maintenance mode. tsdown is built on Rolldown and is typically faster for library builds, with a largely drop-in config surface.
@@ -43,7 +39,3 @@ export default defineConfig({
 
 - `npx tsdown-migrate`
 - `npx skills add rolldown/tsdown --skill tsdown-migrate`
-
-### Requirements
-
-- Node.js >= 20.19 (current active LTS line)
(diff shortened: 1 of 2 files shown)


docs/modules/tsup.md
```diff
@@ -0,0 +1,49 @@
+---
+description: Modern alternative to tsup for bundling TypeScript libraries
+---
+
+# Replacements for `tsup`
+
+## Scope
+
+This replacement is for **existing `tsup` users** only. It is not a recommendation for projects on other bundlers, or for projects that do not need a bundler.
```
43081j: we may not need to specify this, as it could do the reverse of what you intended - it could make it sound like we _recommend against_ using it for other projects.

i'd probably just drop the scope section and hope the title is clear enough that this is specific to tsup.

docs/modules/tsup.md
```diff
@@ -0,0 +1,49 @@
+---
+description: Modern alternative to tsup for bundling TypeScript libraries
```
gameroman, line 2:
```suggestion
description: Modern alternative to the tsup package for bundling TypeScript libraries
```

docs/modules/tsup.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+### Automated migration
+
+- `npx tsdown-migrate`
+- `npx skills add rolldown/tsdown --skill tsdown-migrate`
+
+### Requirements
+
+- Node.js >= 20.19 (current active LTS line)
```
gameroman: This is incorrect, and also I don't think this should be here

gameroman (changes requested):

docs/modules/tsup.md
dreyfus92: happy to tweak if you prefer, but a few docs in the repo omit "the ... package" e.g. `depcheck`, `js-yaml` and this page only documents one replacement so i kept "alternatives" singular.

docs/modules/tsup.md
dreyfus92: good catch 👀

dreyfus92:
thank you @43081j & @gameroman for your reviews ✌🏻

ghostdevv (dismissed):

docs/modules/tsup.md
```diff
@@ -0,0 +1,41 @@
+---
+description: Modern alternative to tsup for bundling TypeScript libraries
+---
+
+# Replacements for `tsup`
+
+## `tsdown`
+
+[tsdown](https://github.com/rolldown/tsdown) is the successor recommended by [tsup's own README](https://github.com/egoist/tsup) since August 2025. The project is effectively in maintenance mode. tsdown is built on Rolldown and is typically faster for library builds, with a largely drop-in config surface.
```
gameroman: Do we need the `since August 2025` part?

43081j: Probably not, I'm happy either way. @dreyfus92 let me know and I'll merge once you've looked

dreyfus92: i've removed it 😄, thank you both ✌🏻

Issue #245, opened 2025-09-19
object-hash


joaopedrodcf:
Object-hash is a library to generate hash from objects,it's no update for 4 years

There are alternatives that are currently maintained:

ohash - https://github.com/unjs/ohash

Ohash is based on object-hash for the object-serialization

PR #749, opened 2026-06-24
feat(preferred): add object-identity as an object-hash replacement

maraisr:
### 🔗 Linked issue

There isn't a dedicated open issue for this one, it builds on the existing `object-hash` replacement page (prior context in #245). Happy to open a tracking issue first if you'd rather discuss it there.

### 📚 Description

This PR suggests [`object-identity`](https://github.com/maraisr/object-identity) as another replacement option for `object-hash`.

- adds `object-identity` to the `object-hash` mapping in `preferred.json` (as a `documented` replacement, alongside `ohash`), and
- updates the examples to use it as the stable serializer.

Full disclosure: I'm the author of `object-identity`, so please take this as you will. I've tried to keep every claim backed by numbers and to be honest about the trade-offs, very happy to adjust or drop anything.

#### Why it fits here

`object-identity` canonicalizes any value into a stable identity string, exactly the lib you reach for when you need a deterministic key to hash or to compare. It's deep and cycle-safe by default, and it's tiny.

**Smaller** — both are zero-dependency; minified + gzipped:

| package | min + gzip |
| --- | --- |
| `object-identity` | **~555 B** |
| `safe-stable-stringify` | ~2.47 KB |

That's roughly 4.5× smaller.

**Faster** — I am using my benchmarks for these two libraries (ops/sec, higher is better):

| workload | `object-identity` | `safe-stable-stringify` |
| --- | --- | --- |
| simple | **910,961** | 593,432 |
| deep | **115,162** | 73,438 |
| deep circular | **175,990** | 138,163 |
| big | 7,095 | **7,377** |
| leafy | **43,393** | 20,510 |

So it's ahead on most cases (≈1.3×–2.1×) and about line-ball on the wide "big" object, where `safe-stable-stringify` is marginally quicker (room for improvement).

> I took heavy inspiration from `safe-stable-stringify`'s test-suite, and pass all of them for their default configuration. So this is very much a drop-in replacement. 

#### Trade-offs / what it doesn't do

`object-identity` is deliberately opinionated: it always produces a canonical, cycle-safe identity and exposes no options, so it doesn't cover some of `safe-stable-stringify`'s options, for example:

- a custom key comparator / toggling `deterministic`
- customising the circular-reference value (or throwing on cycles)
- `maximumDepth` / `maximumBreadth` limits
- a `replacer`, indentation / pretty-printing, and BigInt handling

It also emits an internal identity string rather than valid, human-readable JSON. In practice, for what this utility is usually used for, is for deriving a stable key to feed a hash, or checking structural equality. None of those tend to matter; you just want the same input to map to the same output. If you specifically need canonical JSON out, a `replacer`, or pretty output, `safe-stable-stringify` remains the better tool, which is why it stays alongside.

Very open to feedback, and happy to reword, or keep this docs-only and drop the manifest entry if that sits better with how you'd like to position it. Thanks for considering it, and for maintaining this ❤️

### DIFF
--- a/docs/modules/object-hash.md
+++ b/docs/modules/object-hash.md
@@ -18,6 +18,22 @@ const h = objectHash(obj) // [!code --]
 const h = hash(obj) // [!code ++]
 ```
 
+## `object-identity`
+
+[`object-identity`](https://github.com/maraisr/object-identity) canonicalizes any value into a stable identity string. It's deep, cycle-safe, and dependency-free, and it's a great fit for deriving a deterministic key or comparing structural equality. For most uses of `object-hash`, it should serve you well.
+
+It returns a stable identity rather than a cryptographic hash. If you specifically need a hash digest, follow one of the examples below.
+
+Example:
+
+```ts
+import objectHash from 'object-hash' // [!code --]
+import { identify } from 'object-identity' // [!code ++]
+
+const id = objectHash(obj) // [!code --]
+const id = identify(obj) // [!code ++]
+```
+
 ## Web Crypto
 
 Use the standard [`SubtleCrypto.digest`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) available in modern runtimes. Pair it with a stable serializer (e.g., [`safe-stable-stringify`](https://github.com/BridgeAR/safe-stable-stringify), [`object-identity`](https://github.com/maraisr/object-identity)) to ensure deterministic key ordering.
--- a/docs/modules/object-hash.md
+++ b/docs/modules/object-hash.md
@@ -20,16 +20,16 @@ const h = hash(obj) // [!code ++]
 
 ## Web Crypto
 
-Use the standard [`SubtleCrypto.digest`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) available in modern runtimes. Pair it with a stable serializer (e.g., [`safe-stable-stringify`](https://github.com/BridgeAR/safe-stable-stringify)) to ensure deterministic key ordering.
+Use the standard [`SubtleCrypto.digest`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) available in modern runtimes. Pair it with a stable serializer (e.g., [`safe-stable-stringify`](https://github.com/BridgeAR/safe-stable-stringify), [`object-identity`](https://github.com/maraisr/object-identity)) to ensure deterministic key ordering.
 
 Example:
 
 ```ts
 import objectHash from 'object-hash' // [!code --]
-import stringify from 'safe-stable-stringify' // [!code ++]
+import { identify } from 'object-identity' // [!code ++]
 
 const h = objectHash(obj, { algorithm: 'sha256' }) // [!code --]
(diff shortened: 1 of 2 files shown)


docs/modules/object-hash.md
```diff
@@ -18,18 +18,34 @@ const h = objectHash(obj) // [!code --]
...
+import { identify } from 'object-identity' // [!code ++]
+
+const id = objectHash(obj) // [!code --]
+const id = identify(obj) // [!code ++]
+```
+
 ## Web Crypto
 
-Use the standard [`SubtleCrypto.digest`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) available in modern runtimes. Pair it with a stable serializer (e.g., [`safe-stable-stringify`](https://github.com/BridgeAR/safe-stable-stringify)) to ensure deterministic key ordering.
+Use the standard [`SubtleCrypto.digest`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) available in modern runtimes. Pair it with a stable serializer (e.g., [`safe-stable-stringify`](https://github.com/BridgeAR/safe-stable-stringify), [`object-identity`](https://github.com/maraisr/object-identity)) to ensure deterministic key ordering.
```
gameroman:
```suggestion
Use the standard [`SubtleCrypto.digest`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) available in modern runtimes. Pair it with a stable serializer (e.g., [`safe-stable-stringify`](https://github.com/BridgeAR/safe-stable-stringify), or [`object-identity`](https://github.com/maraisr/object-identity)) to ensure deterministic key ordering.
```

Or maybe even just

```suggestion
Use the standard [`SubtleCrypto.digest`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) available in modern runtimes. Pair it with a stable serializer (e.g., [`object-identity`](https://github.com/maraisr/object-identity)) to ensure deterministic key ordering.
```

maraisr: Great suggestion, Ive applied this.

dreyfus92 (commented):
hey @maraisr thanks for this, and for the honest write-up of the trade-offs & the disclosure are appreciated.

happy to take it as a documented option, but a few changes first:

- revert the safe-stable-stringify swaps in the web crypto + bun examples. the pr says it "stays alongside" but the diff removes it; let's leave those examples on safe-stable-stringify and let object-identity stand on its own section.
- order it after ohash in the object-hash `replacements` array. this isn't cosmetic: the cli picks the first compatible entry, so prepending it makes `object-identity` the surfaced recommendation over ohash, and replacements.fyi renders the array in order so it'd also be top-listed there. given ohash is the maintained successor at ~1000x the adoption, it should lead.

worth flagging for transparency that `object-identity` is ~6.5k/week vs our ~10k soft bar, so i'm taking it as documented-but-not-leading rather than promoting it ahead of ohash.

also i'd love @43081j's input before merging this one.

maraisr:
Thank you for your review here @dreyfus92, and do apologise for any inconvenience this may have caused you. Maybe another time!

43081j:
Very happy to add this once you see some adoption 👍

Please do recreate this PR then as it would be nice to include it

PR #751, opened 2026-06-24
feat: expose util subpath

TheAlexLichter:
### 🔗 Linked issue

None (Perf related)

### 📚 Description

This PR adds a stable `module-replacements/util` export for `resolveDocUrl`, avoiding the root import that pulls manifests into client bundles.

This results in a smaller bundle if you dont need the manifest, saving ~26kB gzipped.

### Alternatives

Use `dist/utils.js` which works today but does not feel like it part of the public API contract.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,12 @@
         "default": "./dist/main.js"
       }
     },
+    "./util": {
+      "import": {
+        "types": "./dist/util.d.ts",
+        "default": "./dist/util.js"
+      }
+    },
     "./dist/util.js": {
       "import": {
         "types": "./dist/util.d.ts",


gameroman:
You can already do that via `./dist/util.js`

```js
import { ... } from 'module-replacements/dist/util.js' 
```

TheAlexLichter:
> You can already do that via `./dist/util.js`
> 
> ```js
> import { ... } from 'module-replacements/dist/util.js' 
> ```

True, it doesn't hold anyone back using it already. But relying on a `dist/*` path always feels a bit "unofficial" because it is coupled to how the build path looks like.

dreyfus92 (commented):
hey @TheAlexLichter ✌🏻

did a bit of measuring before this grows the export surface, since the manifests only tag along in one specific case.
 
bundled `import { resolveDocUrl } from 'module-replacements'` (root) vs a lean subpath, minified:
 
| | root import | subpath |
|---|---|---|
| **esbuild**, no `sideEffects` | ~167 kb | ~300 b |
| **esbuild**, `sideEffects: false` | ~300 b | ~300 b |
| **rollup** (either way) | ~500 b | ~500 b |
 
the manifests only get pulled in under **esbuild when this package doesn't declare `sideEffects`**. rollup — what npmx ships in prod — already tree-shakes them out of the root import today. and root already re-exports `resolveDocUrl` via `export * from './util.js'`, so it's importable from the entrypoint right now.
 
so i think the smaller fix is adding `"sideEffects": false` to **this repo's** `package.json` (module-replacements), not a new export path. that's a declaration the package makes about itself. consumers' bundlers read it to prune our unused manifest code, so it's one line here that fixes every lean import downstream, no consumer config needed. the package is pure (json data + pure fns, no top-level side effects), so the flag is accurate. with it, the root import drops to ~300 b under esbuild too, matching what a subpath would buy and it'd let us eventually retire the `./dist/util.js` export instead of adding a sibling alongside it.

open to being wrong, is there a consumer/bundler combo that ignores a dependency's `sideEffects` where the subpath is the only thing that gets the win?

wdyt @43081j?

TheAlexLichter:
Good call @dreyfus92 🙌 

`sideEffects: false` should be even better. we should retire dist/utils.js once there is a major version in the future 👍

PR #755, opened 2026-06-26
fix: move define-properties to preferred manifest

morgan-coded:
### 🔗 Linked issue

https://github.com/es-tooling/eslint-plugin-depend/issues/35

### 📚 Description

Moves `define-properties` from the native manifest to the preferred manifest because it is not always a direct native drop-in for `Object.defineProperties`.

This also adds a docs page for the preferred entry, with the distinction that `Object.defineProperties` is enough when callers only need to define descriptors, while `define-properties` has extra behavior around existing properties and predicate-controlled overrides.

Checks:
`pnpm lint`

### DIFF
--- a/docs/modules/define-properties.md
+++ b/docs/modules/define-properties.md
@@ -11,22 +11,20 @@ description: Native alternatives to the define-properties package for defining m
 Example:
 
 ```ts
-import defineProperties from 'define-properties'
+import defineProperties from 'define-properties' // [!code --]
 
 const obj = {}
-
-defineProperties(obj, {
-  enabled: true
-})
-
-Object.defineProperties(obj, {
+const descriptors = {
   enabled: {
     configurable: true,
     enumerable: false,
     value: true,
     writable: false
   }
-})
+}
+
+defineProperties(obj, { enabled: true }) // [!code --]
+Object.defineProperties(obj, descriptors) // [!code ++]
 ```
 
 `define-properties` also skips existing properties unless a matching predicate returns `true`. If your code relies on that behavior, keep the condition outside the descriptor map when migrating:
--- a/docs/modules/define-properties.md
+++ b/docs/modules/define-properties.md
@@ -1,0 +1,47 @@
+---
+description: Native alternatives to the define-properties package for defining multiple object properties
+---
+
+# Replacements for `define-properties`
+
+## `Object.defineProperties` (native)
+
+[`Object.defineProperties`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperties) can define multiple properties on an object without an extra dependency.
+
+Example:
+
+```ts
+import defineProperties from 'define-properties'
+
+const obj = {}
+
+defineProperties(obj, {
+  enabled: true
+})
+
+Object.defineProperties(obj, {
+  enabled: {
+    configurable: true,
+    enumerable: false,
+    value: true,
+    writable: false
+  }
+})
+```
+
+`define-properties` also skips existing properties unless a matching predicate returns `true`. If your code relies on that behavior, keep the condition outside the descriptor map when migrating:
+
+```ts
+const descriptors = {
+  enabled: {
+    configurable: true,
+    enumerable: false,
+    value: true,
+    writable: false
+  }
+}
+
+if (!Object.hasOwn(obj, 'enabled') || shouldOverrideEnabled()) {
+  Object.defineProperties(obj, descriptors)
+}
+```
--- a/docs/modules/define-properties.md
+++ b/docs/modules/define-properties.md
@@ -30,6 +30,8 @@ Object.defineProperties(obj, descriptors) // [!code ++]
 `define-properties` also skips existing properties unless a matching predicate returns `true`. If your code relies on that behavior, keep the condition outside the descriptor map when migrating:
 
 ```ts
+import defineProperties from 'define-properties' // [!code --]
+
 const descriptors = {
   enabled: {
     configurable: true,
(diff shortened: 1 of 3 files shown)


gameroman:
A package goes to `preferred` manifest if it has a documented replacements so you would need to add a .md file explaining the replacement,
for example how it's done for `buffer-equal-constant-time`


Otherwise it should stay in the `native` manifest since it is still replaceable by native functionality

gameroman:
Also please use the PR template

43081j:
given that you can conditionally define properties natively, too, can you explain why someone should use a package to do it instead?

e.g.

```ts
Object.defineProperties(obj, {
  // Always defined
  id: {
    value: 303
  },

  // Conditionally defined
  ...(flags.enableFoo && {
    foo: {
      value: "foo value"
    },
  }),

  ...(flags.enableBar && {
    bar: {
      value: "bar value"
    },
  }),
});
```

gameroman:
@43081j I might be wrong but this pr honestly looks like it was opened by a bot

morgan-coded:
I updated this to follow the preferred-entry convention: `define-properties` now has a docs page and the manifest entry links to it.

I’m not trying to argue that `Object.defineProperties` can’t cover conditional cases with normal JS. The issue is that the current native warning feels too absolute for this package, since `define-properties` also has behavior around existing properties and predicate-controlled overrides. The docs now frame `Object.defineProperties` as the right replacement when that extra behavior is not needed.

Happy to switch direction or close this if you’d rather keep the native classification.

docs/modules/define-properties.md
```diff
@@ -0,0 +1,47 @@
+---
+description: Native alternatives to the define-properties package for defining multiple object properties
+---
+
+# Replacements for `define-properties`
+
+## `Object.defineProperties` (native)
+
+[`Object.defineProperties`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperties) can define multiple properties on an object without an extra dependency.
+
```
gameroman, line 10: You need to show a diff for migration like in https://github.com/e18e/module-replacements/blob/main/docs/modules/clipboardy.md for example

gameroman (changes requested):

morgan-coded:
Updated the example to use the migration diff format. `pnpm lint` is still passing.

docs/modules/define-properties.md
```diff
@@ -0,0 +1,45 @@
...
+    enumerable: false,
+    value: true,
+    writable: false
+  }
+}
+
+if (!Object.hasOwn(obj, 'enabled') || shouldOverrideEnabled()) {
+  Object.defineProperties(obj, descriptors)
+}
+```
```
gameroman, line 49: Could you add a diff for this too please?

morgan-coded:
Updated the second example to use the migration diff format too. `pnpm lint` is still passing.

43081j:
> I’m not trying to argue that `Object.defineProperties` can’t cover conditional cases with normal JS. The issue is that the current native warning feels too absolute for this package, since `define-properties` also has behavior around existing properties and predicate-controlled overrides. The docs now frame `Object.defineProperties` as the right replacement when that extra behavior is not needed.

If you understand that `Object.defineProperties` _can_ cover conditional cases, what is the extra behaviour we _can't_ do natively that justifies pulling a package in?

You say `also has behavior around existing properties and predicate-controlled overrides`. Given we know you can do conditional properties natively, that just leaves `behavior around existing properties`, yes? What is that behaviour that can't be done natively?

morgan-coded:
Thanks for pushing on the distinction here.

I rechecked `define-properties`, and I agree this is more convenience/ergonomics than a real native capability gap. The existing behavior around skipping existing keys and predicate-controlled overrides is useful for shim-style code, but it can still be expressed with `Object.defineProperties` and normal JS conditionals.

So I think `native` is the better classification after all. I’m going to close this to avoid churn.

ljharb:
For the record, I don't care if e18e wants to discourage use of `define-properties` or not, but in no way whatsoever is it a polyfill or replacement for `Object.defineProperties`, despite a relatively popular blog post whose premise is based on that flawed understanding.

43081j:
it is not a polyfill but does feature-detect and fall back when `Object.defineProperties` doesn't exist.

other than that, it seems it is a convenience wrapper around `Object.defineProperties` if you prefer that API over the native one. 

that's why it is a native replacement. the same functionality can be achieved if you're willing to change API shape and don't care about very old engines. that doesn't mean the package is unnecessary in general - some people may just prefer that API shape.

ljharb:
No, it does zero feature detection of Object.defineProperties - only around `Object.defineProperty` via the `define-data-property` package.

43081j:
sorry, i meant with the functionality of its dependencies. it does feature detection of `Object.defineProperty` then, via its dependencies, yes?

so like i mentioned above, it does feature-detect and fallback, and provides a convenience wrapper. neither of which justifies its use in this scope since we're targeting engines that don't need the fallback, and users who don't need a different API shape.

ljharb:
One of its _dependencies_ does, yes, but _it_ does not - and replacing either of these with JS (as any package can be) will save you virtually zero bytes, due to the API shape. There's about 5 lines of code that falls back when it's not available - all the rest is necessary in every engine, no matter how modern.

In other words, discourage whichever packages you like, but this is simply not a "convenience wrapper" around a native API, it's something that uses native APIs and adds a negligible amount of fallback code to support ES3 engines.

43081j:
Necessary to have the API shape/convenience wrapper it gives. Since we don't need that or the fallback logic in this case, it can be replaced by the built in 👍

As I mentioned above, it provides backwards compat and a convenience wrapper. Not concerned with if a dependency is responsible or not - it and the dependencies it pulls in results in providing a backwards compat fallback on top of the convenience wrapper. 

Neither of which are needed in this case. All good, if someone prefers that wrapper or needs the backwards compat, those are the only reasons to use it.

ljharb:
I think we're on the same page, thanks :-)

PR #780, opened 2026-07-06
feat: add sanitize-html → neosanitize replacement

### LINKED ISSUE #781: [Replacement]: neosanitize
### Package to replace

sanitize-html

### Suggested replacement(s)

neosanitize package. By default, we suggest `neosanitize/legacy` as drop-in, and `neosanitize` with the prebuilt whatwg/parse5/htmlparser for more secure/correct as the correct path

### Manifest type

### Rationale

sanitize-html ships entirity of postcss and other libraries just for very simple parsing at many places. neosanitize/legacy is a no-dep drop-in replacement passing all the tests of sanitize-html, and the relevant test suites of the upstream deps.

neosanitize itself comes by default a very fast WHATWG-parser, which is swappable with parse5, htmlparser2 as well

### Availability

_No response_

### Code example (optional)

```js

```

PuruVJ:
## Summary

Adds `sanitize-html` to the `preferred` manifest, with:

- **`neosanitize/legacy`** as a direct, drop-in replacement — a byte-identical port of `sanitize-html` 2.x (same API, same options, same output), so migration is just an import swap.
- **`neosanitize`** (root module) as the recommended alternative — a fast, browser-faithful WHATWG parser (100% html5lib tokenizer conformance) with a deny-by-default policy, using a builder-pattern API.

Includes a `docs/modules/sanitize-html.md` page with migration examples for both engines.

## Migration

```ts
// Zero-change drop-in
import sanitizeHtml from 'sanitize-html'      // before
import sanitizeHtml from 'neosanitize/legacy' // after
```

## Notes

- Placed in `preferred` (performance/footprint replacement — not native, not a micro-utility).
- Both replacements are represented as `documented` entries; the mapping points to the new e18e doc page.
- `pnpm lint` passes (JSON schema, sort order, doc-path checks, Prettier).
Closes #781

### DIFF
--- a/docs/modules/sanitize-html.md
+++ b/docs/modules/sanitize-html.md
@@ -1,0 +1,56 @@
+---
+description: Replacing sanitize-html with neosanitize, a zero-dependency, isomorphic HTML sanitizer with a drop-in legacy engine and a faster WHATWG-based engine
+---
+
+# Replacements for `sanitize-html`
+
+[`sanitize-html`](https://github.com/apostrophecms/sanitize-html) is a widely used HTML sanitizer, but it pulls in a large dependency tree (including `htmlparser2`, `parse-srcset`, `postcss`, and others) and only runs in Node-like environments.
+
+[`neosanitize`](https://github.com/PuruVJ/neosanitize) is a zero-dependency, isomorphic (works in Node, browsers, and edge runtimes) HTML sanitizer written in TypeScript. It ships two engines:
+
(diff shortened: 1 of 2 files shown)


PuruVJ:
Retargeting to my fork instead.

gameroman (changes requested):
I think we need an issue with the template first

PuruVJ:
Socket IO already using this lib internally, if the 10k rule isn't a hard-limit for usage: https://x.com/puruvjdev/status/2074181968624066813

PR #811, opened 2026-07-13
feat(micro-utilities): add is-set to replacements

### LINKED ISSUE #713: [Replacement]: `is-set`
### Package to replace

`is-set`

### Suggested replacement(s)

```js
Object.prototype.toString.call(v) === "[object Set]"
```

### Manifest type

dreyfus92:
### 🔗 Linked issue

closes: #713

### 📚 Description

adds `is-set` to replacements

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -258,6 +258,12 @@
       "description": "You can use `instanceof RegExp` to check if a value is a regular expression, or if cross-realm, use `Object.prototype.toString.call(v) === \"[object RegExp]\"`.",
       "example": "const isRegExp = (v) => v instanceof RegExp;\n// for cross-realm\nconst isRegExpCrossRealm = Object.prototype.toString.call(v) === \"[object RegExp]\";"
     },
+    "snippet::is-set": {
+      "id": "snippet::is-set",
+      "type": "simple",
+      "description": "You can use `instanceof Set`, or if cross-realm, use `Object.prototype.toString.call(v) === \"[object Set]\"`",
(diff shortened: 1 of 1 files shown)


manifests/micro-utilities.json
```diff
@@ -258,6 +258,12 @@
       "description": "You can use `instanceof RegExp` to check if a value is a regular expression, or if cross-realm, use `Object.prototype.toString.call(v) === \"[object RegExp]\"`.",
       "example": "const isRegExp = (v) => v instanceof RegExp;\n// for cross-realm\nconst isRegExpCrossRealm = Object.prototype.toString.call(v) === \"[object RegExp]\";"
     },
+    "snippet::is-set": {
+      "id": "snippet::is-set",
+      "type": "simple",
+      "description": "You can use `instanceof Set`, or if cross-realm, use `Object.prototype.toString.call(v) === \"[object Set]\"`",
```
```suggestion
      "description": "You can use `instanceof Set` to check if a value is a `Set`, or if cross-realm, use `Object.prototype.toString.call(v) === \"[object Set]\"`",
```

gameroman (changes requested):

manifests/micro-utilities.json
dreyfus92: i think that's redundant wouldn't it? user should know that a priori 😄

gameroman: Well, otherwise it just repeats what is already in the snippet

Also this would match what we already have for `is-regex`

I'm not sure if there is a simpler way to do it

43081j: i think its fine either way but ill merge the change just so its consistent

43081j (approved):

PR #815, opened 2026-07-16
docs(dotenv): suggest process.loadEnvFile as native replacement

### LINKED ISSUE #620: [dotenv] Additionally suggest process.loadEnvFile(path)
[dotenv](https://github.com/e18e/module-replacements/blob/main/docs/modules/dotenv.md) module replacement would be nice also to mention the [process.loadEnvFile(path)](https://nodejs.org/api/process.html#processloadenvfilepath) Added in: v21.7.0, v20.12.0

unrevised6419:
## What

Adds [`process.loadEnvFile(path)`](https://nodejs.org/docs/latest-v20.x/api/process.html#processloadenvfilepath) (Node.js v20.12.0+) as the first suggested native replacement in the `dotenv` docs.

It is the closest drop-in for `dotenv.config()` when env vars must be loaded programmatically rather than via the `--env-file` flag.

## Verified

Behavior claims checked empirically on Node v24.16.0:

- Throws `ENOENT` when the file is missing.
- Defaults to `.env` in the current working directory when called with no path.

Closes #620

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/modules/dotenv.md
+++ b/docs/modules/dotenv.md
@@ -4,9 +4,9 @@ description: Native Node.js alternatives to the dotenv package for loading and m
 
 # Replacements for `dotenv`
 
-## Node.js `process.loadEnvFile`
+## `process.loadEnvFile` (native, since Node.js v20.12.0)
 
-[`process.loadEnvFile`](https://nodejs.org/docs/latest-v20.x/api/process.html#processloadenvfilepath) (Node.js v20.12.0+) loads a `.env` file into `process.env` at runtime. It is the closest drop-in replacement for `dotenv.config()` when you need to load environment variables programmatically instead of via the `--env-file` flag.
+[`process.loadEnvFile`](https://nodejs.org/api/process.html#processloadenvfilepath) (Node.js v20.12.0+) loads a `.env` file into `process.env` at runtime.
 
 It throws if the file is missing. Defaults to `.env` in the current working directory when called without a path.
 
--- a/docs/modules/dotenv.md
+++ b/docs/modules/dotenv.md
@@ -4,6 +4,19 @@ description: Native Node.js alternatives to the dotenv package for loading and m
 
 # Replacements for `dotenv`
 
+## Node.js `process.loadEnvFile`
+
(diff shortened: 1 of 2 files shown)


docs/modules/dotenv.md
```diff
@@ -4,6 +4,19 @@ description: Native Node.js alternatives to the dotenv package for loading and m
 
 # Replacements for `dotenv`
 
+## Node.js `process.loadEnvFile`
```
gameroman:
```suggestion
## `process.loadEnvFile` (native, since Node.js v20.12.0)
```

docs/modules/dotenv.md
```diff
@@ -4,6 +4,19 @@ description: Native Node.js alternatives to the dotenv package for loading and m
...
+
+[`process.loadEnvFile`](https://nodejs.org/docs/latest-v20.x/api/process.html#processloadenvfilepath) (Node.js v20.12.0+) loads a `.env` file into `process.env` at runtime. It is the closest drop-in replacement for `dotenv.config()` when you need to load environment variables programmatically instead of via the `--env-file` flag.
```
gameroman:
```suggestion
[`process.loadEnvFile`](https://nodejs.org/api/process.html#processloadenvfilepath) (Node.js v20.12.0+) loads a `.env` file into `process.env` at runtime.
```

gameroman (changes requested):
Also add `process.loadEnvFile` to the manifest please

unrevised6419:
Done — added `process.loadEnvFile` as a native replacement entry and listed it for the `dotenv` mapping. Manifest validation + prettier pass.

dreyfus92:
@gameroman if this is a native replacement, shouldn't this live within the native manifest instead of preferred?

gameroman:
> @gameroman if this is a native replacement, shouldn't this live within the native manifest instead of preferred?

Do you mean `process.loadEnvFile` or the whole thing?

I think it's in the preferred manifest because it has docs, and previously we didn't have node-native replacements

dreyfus92 (approved):

43081j:
@unrevised6419 before we merge this, can you come say Hi so we can interact with you rather than your agent.

these days it is important we know we're dealing with a person since some of these things lead to rep farming etc.

unrevised6419:
Hi @43081j 😄 The agent is under my assistance, it's not on his own.

PR #816, opened 2026-07-17
fix: correct `js-yaml` maintenance status in intro.

### LINKED ISSUE #813: js-yaml documentation incorrect
`js-yaml` documentation indicates the package is not maintained. However, a new major version was released [July 1, 2026](https://www.npmjs.com/package/js-yaml/v/5.2.1). This may or may not change the recommendations, but the note about being unmaintained should be corrected regardless.

https://github.com/e18e/module-replacements/blob/main/docs/modules/js-yaml.md?plain=1#L7

dreyfus92:
### 🔗 Linked issue

closes: #813

### 📚 Description

updates `js-yaml` doc updating maintenance status in intro.

### DIFF
--- a/docs/modules/js-yaml.md
+++ b/docs/modules/js-yaml.md
@@ -4,7 +4,7 @@ description: Modern alternatives to js-yaml for YAML parsing and stringifying
 
 # Replacements for `js-yaml`
 
-`js-yaml` appears to be unmaintained and has known spec-compliance issues.
+`js-yaml` is actively maintained. Alternatives below may still be preferable for comment/AST support, zero dependencies, or Bun-native usage.
 
 ## `yaml`
 
--- a/docs/modules/js-yaml.md
+++ b/docs/modules/js-yaml.md
@@ -4,7 +4,7 @@ description: Modern alternatives to js-yaml for YAML parsing and stringifying
 
 # Replacements for `js-yaml`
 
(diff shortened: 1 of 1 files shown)


docs/modules/js-yaml.md
```diff
@@ -4,7 +4,7 @@ description: Modern alternatives to js-yaml for YAML parsing and stringifying
 
 # Replacements for `js-yaml`
 
-`js-yaml` appears to be unmaintained and has known spec-compliance issues.
+`js-yaml` is actively maintained. Alternatives below may still be preferable for comment/AST support, zero dependencies, or Bun-native usage.
```
gameroman: I'd just remove this completely

```suggestion
```

dreyfus92: fair on not asserting maintenance status, it'll just rot and we'll end up back here with another issue. but i'd rather not strip the intro entirely, since a reader landing here with no context reads the alternatives as "migrate now, this is dead." how about we drop the status claim and frame it around what the alternatives give you instead:

> "The alternatives below can offer better spec compliance, comment/AST support, zero dependencies, or native Bun support."

keeps it accurate, no claim that ages. wdyt?

gameroman: I'd just mention "better spec compliance" and "smaller install size"

joaopedrodcf (approved):

PR #817, opened 2026-07-18
Add fastify to preferred replacements for Express

mcollina:
👋 I think Fastify should be on this list.

### DIFF
--- a/docs/modules/express.md
+++ b/docs/modules/express.md
@@ -83,3 +83,23 @@ app.get('/', () => 'Hello world') // [!code ++]
 
 app.listen(3000)
 ```
+
+## `fastify`
+
+[`fastify`](https://fastify.dev/) is a fast, low-overhead web framework for Node.js with a powerful plugin architecture and schema-based validation and serialization.
+
+Example:
+
+```ts
+import express from 'express' // [!code --]
+import Fastify from 'fastify' // [!code ++]
+
+const app = express() // [!code --]
+const app = Fastify() // [!code ++]
+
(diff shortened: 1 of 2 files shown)


gameroman:
I think the reason why we didn't add it in the first place was that, it has the biggest install size and almost as many deps as express

Let's see what others think, cc @43081j @dreyfus92

mcollina:
The goal of a framework is not to be small on disk, or have a lot or few dependencies, its to provide the features its users needs.

Size on disk is almost irrelevant for servers.

---

Last but not least, I would recommend against keeping this page at all. A lot of people spends significant amount of time maintaining frameworks - it's a lot of work. Entire companies depend on them, and funding this work is complicated.

This page set a precedent of "picking a winner" between frameworks, and its equivalent of naming React the replacement of Angular.


While I think the initiative tackling small modules and tiny deps is worthwhile, Express should not be on this list.

43081j:
> The goal of a framework is not to be small on disk, or have a lot or few dependencies, its to provide the features its users needs.
> 
> Size on disk is almost irrelevant for servers.

For another time, but this kind of dismissal of package size is a large part of why the ecosystem is so bloated today. The sum of all the people having a similar attitude towards it results in huge amounts of wasted bandwidth and storage.

It'll do the ecosystem a lot of good for you to learn that size on disk and bandwidth does indeed matter - always.

Given you've opened a follow-up/related issue in npmx about this same thing, I will reply there instead and leave this until that one is resolved. A side note, for future reference, trying to drive engagement to your issue through socials will only result in noise rather than progress in most cases.

mcollina:
Closing.

PR #829, opened 2026-07-21
feat(micro-util): path-key

### LINKED ISSUE #828: [Replacement]: path-key
### Package to replace

path-key

### Suggested replacement(s)

``` js

const pathKey = Object.keys(env).findLast(k => k.toUpperCase() === 'PATH') ?? 'PATH'

```

### Manifest type

### Rationale

path-key - 256M downloads/week

On non-Windows, it returns PATH. 
On Windows, it searches your process.env for "PATH" case-insensitive. 

I wonder who these people are with a PAth or a patH

For what it's worth, if you need to _set_ the same key, it has some use. But still, you can basically one-liner it:

Object.keys(env).findLast(k => k.toUpperCase() === 'PATH') ?? 'PATH'

Would love to know if people think this is still worthy of a package

from
https://mu.social/profile/43081j.com/post/3mr66paffl225

### Availability

All modern browsers (baseline since August 2022) and node 18+ 

### Code example (optional)

```js
fucntion pathKey() {
return Object.keys(env).findLast(k => k.toUpperCase() === 'PATH') ?? 'PATH'

}
```

MaxwellCohen:
### 🔗 Linked issue
Fixes: #828

### 📚 Description
Adds `pathKey` micro utility to replace `path-key` depency

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -358,7 +358,7 @@
       "id": "snippet::path-key",
       "type": "simple",
       "description": "You can use `Object.keys` with `Array.prototype.findLast` to find the case-insensitive `PATH` environment variable key.",
-      "example": "const pathKey = (env = process.env) => Object.keys(env).findLast((k) => k.toUpperCase() === 'PATH') ?? 'PATH';"
+      "example": "const pathKey = Object.keys(process.env).findLast((k) => k.toUpperCase() === 'PATH') ?? 'PATH';"
     },
     "snippet::regexp-copy": {
       "id": "snippet::regexp-copy",
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -354,6 +354,12 @@
       "description": "You can use `Object.entries` and `Array.prototype.reduce`.",
       "example": "const objectReduce = (obj, fn, initial) => Object.entries(obj).reduce((acc, [k, v]) => fn(acc, k, v), initial);"
     },
+    "snippet::path-key": {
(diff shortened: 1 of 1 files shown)


manifests/micro-utilities.json
```diff
@@ -354,6 +354,12 @@
       "description": "You can use `Object.entries` and `Array.prototype.reduce`.",
       "example": "const objectReduce = (obj, fn, initial) => Object.entries(obj).reduce((acc, [k, v]) => fn(acc, k, v), initial);"
     },
+    "snippet::path-key": {
+      "id": "snippet::path-key",
+      "type": "simple",
+      "description": "You can use `Object.keys` with `Array.prototype.findLast` to find the case-insensitive `PATH` environment variable key.",
+      "example": "const pathKey = (env = process.env) => Object.keys(env).findLast((k) => k.toUpperCase() === 'PATH') ?? 'PATH';"
```
43081j: i'd probably leave it as a constant value since the common use case will be against `process.env`. i.e. `pathKey` doesn't need to be a function in most cases

gameroman: So like this

```suggestion
      "example": "const pathKey = Object.keys(process.env).findLast((k) => k.toUpperCase() === 'PATH') ?? 'PATH';"
```

MaxwellCohen: @gameroman, thank you for adding this suggestion. I was just about to add this change in about 10 min from now

gameroman (approved):

PR #844, opened 2026-07-30
feat: add Yup replacements

### LINKED ISSUE #822: [Replacement]: yup
# Package to replace

yup

## Suggested replacement(s)

- `valibot` — 0 dependencies, ~0.3-1.9KB min+gzip (tree-shakable), implements [Standard Schema](https://standardschema.dev)
- `zod` (v4) — 0 dependencies, ~17KB full build / ~2-4KB with the tree-shakable "Zod Mini" build, implements Standard Schema, much larger existing ecosystem (react-hook-form resolvers, tRPC, etc.)

## Manifest type

## Rationale

`yup` (~10-11M downloads/week) has 4 direct dependencies (`property-expr`, `tiny-case`, `toposort`, `type-fest`) and a monolithic, chainable-builder architecture that isn't tree-shakable — a typical schema comes to ~20-25KB min+gzip regardless of how much of the API surface is actually used. It has no [Standard Schema](https://standardschema.dev) support, which increasingly matters for interop with libraries like react-hook-form and tRPC that now accept any Standard-Schema-compliant validator. It's still maintained, but release cadence is slow relative to `valibot`/`zod` v4.

Both suggested replacements have 0 dependencies, are meaningfully smaller (with `valibot` tree-shaking down to single-digit-KB for realistic schemas vs. yup's fixed ~20-25KB), and support Standard Schema out of the box.

Comparative bundle-size/perf data across validators: [schemabenchmarks.dev](https://schemabenchmarks.dev/) (note: the site is descriptive — graphs only, no explicit recommendation — the conclusions above are our own synthesis of its data plus npm download/maintenance signals).

## Availability

N/A — not a native replacement. Both `valibot` and `zod` are plain npm packages supporting Node.js and all modern browsers, no special runtime version requirements.

## Code example (optional)

```ts
import * as yup from 'yup' // [!code --]
import * as v from 'valibot' // [!code ++]

const schema = yup.object({ // [!code --]
  name: yup.string().required(), // [!code --]
  age: yup.number().min(0) // [!code --]
}) // [!code --]

const schema = v.object({ // [!code ++]
  name: v.string(), // [!code ++]
  age: v.pipe(v.number(), v.minValue(0)) // [!code ++]
}) // [!code ++]

await schema.validate(input) // [!code --]
v.parse(schema, input) // [!code ++]
```

yslpn:
### 🔗 Linked issue

Resolves #822

### 📚 Description

Adds `yup` to the preferred replacements manifest with two modular, tree-shakable alternatives:

- [`valibot`](https://valibot.dev/)
- [`zod/mini`](https://zod.dev/packages/mini), provided by `zod@^4.0.0`

The shared schema-validation guide is designed for reuse by future schema-library mappings. It documents required fields, empty strings, validation pipelines/checks, async validation, type coercion, unknown object keys, and Zod Mini error locales. Both alternatives support [Standard Schema](https://standardschema.dev/).

The guide links to Zod Mini documentation and Schema Benchmarks without embedding version-sensitive benchmark figures.

### ⚠️ Disclosure

I am a maintainer of Valibot.

### ✅ Validation

- `pnpm lint`
- `node scripts/validate-pr.ts`


docs/modules/yup.md
```diff
@@ -0,0 +1,61 @@
+---
+description: Modern alternatives to the yup package for TypeScript schema validation
```
43081j: i would actually name this file something like `schema-validation.md` i think, and update the description:


```suggestion
description: Modern alternatives for TypeScript schema validation
```

@gameroman what do you think? my reasoning is that we'll inevitably add more schema libraries over time. so this is a similar situation to `fetch` probably

gameroman: Maybe, in that case the whole doc should be changed probably to not be `yup`-specific?

yslpn: @gameroman @43081j I think it's a good idea to make it general. 

If we reach an agreement, I'm ready to rewrite the entire document, making it less specific.

docs/modules/schema-validation.md
```diff
@@ -0,0 +1,49 @@
+---
+description: Modern alternatives for TypeScript schema validation
+---
+
+# TypeScript schema validation
+
+This page contains common, recommended alternatives for TypeScript schema validation. [`valibot`](https://valibot.dev/) and [`zod/mini`](https://zod.dev/packages/mini) are modular, tree-shakable options that both support [Standard Schema](https://standardschema.dev/), allowing them to work with schema-agnostic integrations.
```
43081j:
```suggestion
This page contains common, recommended alternatives for TypeScript schema validation.
```

We explain the individual suggestions below and don't need to repeat them here as the list will grow

43081j (changes requested):

docs/modules/schema-validation.md
yslpn: Thank you, agree

43081j (approved):

PR #890, opened 2026-08-08
feat: add `set-function-length` to replacements

### LINKED ISSUE #710: [Replacement]: `set-function-length`
### Package to replace

`set-function-length`

### Suggested replacement(s)

```js
function fn() {}

Object.defineProperty(fn, 'length', { value: 3 })
```

### Manifest type

### Rationale

native / simple util

### Availability

_No response_

### Code example (optional)

```js

```

gameroman:
### 🔗 Linked issue

Closes #710

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -382,7 +382,7 @@
       "id": "snippet::set-function-length",
       "type": "simple",
       "description": "You can set `length` of a function using `Object.defineProperty`.",
-      "example": "Object.defineProperty(fn, 'length', { value: newLength, configurable: true });"
+      "example": "const setFunctionLength = (fn, length) => Object.defineProperty(fn, 'length', { value: length, configurable: true });"
     },
     "snippet::set-tostringtag": {
       "id": "snippet::set-tostringtag",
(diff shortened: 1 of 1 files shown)


manifests/micro-utilities.json
```diff
@@ -366,6 +366,12 @@
       "description": "You can create a copy of a regular expression using the `RegExp` constructor.",
       "example": "const copyRegExp = (regexp) => new RegExp(regexp);"
     },
+    "snippet::set-function-length": {
+      "id": "snippet::set-function-length",
+      "type": "simple",
+      "description": "You can set `length` of a function using `Object.defineProperty`.",
+      "example": "Object.defineProperty(fn, 'length', { value: newLength, configurable: true });"
```
dreyfus92: small nit, could we match the shape of `snippet::set-tostringtag`? it's written as a reusable arrow, and since `Object.defineProperty` returns the target you also get the same return-the-fn behavior as the actual package for free

```suggestion
      "example": "const setFunctionLength = (fn, length) => Object.defineProperty(fn, 'length', { value: length, configurable: true });"
```

gameroman: Lets see what @43081j thinks

43081j: for consistency it makes sense, but most usages probably don't need to extract it to a function since its a one-liner. the same could be said for other snippets

let's put it in a function for now so it is consistent, but discuss on discord about if any of them should be this shape

43081j (approved):

PR #900, opened 2026-08-09
feat: add cross-spawn replacement

### LINKED ISSUE #887: [Replacement]: `cross-spawn`
### Package to replace

`cross-spawn`

### Suggested replacement(s)

[`tinyexec`](https://main.npmx.dev/package/tinyexec)

### Manifest type

### Rationale

`tinyexec` is smaller, has 0 dependencies and can handle the same job, also exports `normalizeSpawnCommand`

### Availability

_No response_

### Code example (optional)

```js

```

anilloutombam:
### 🔗 Linked issue

Closes #887

### 📚 Description

Maps `cross-spawn` to the existing `tinyexec` replacement.

`tinyexec` provides the same core functionality with a smaller, dependency-free implementation.

### DIFF
--- a/docs/modules/cross-spawn.md
+++ b/docs/modules/cross-spawn.md
@@ -1,0 +1,33 @@
+---
+description: Modern alternatives to the cross-spawn package for spawning child processes
+---
+
+# Replacements for `cross-spawn`
+
+## `tinyexec`
+
+[`tinyexec`](https://github.com/tinylibs/tinyexec) is a minimal, dependency-free process execution library.
+
+`cross-spawn` follows the Node.js `spawn` API and returns a `ChildProcess`. In comparison, `tinyexec` returns an awaitable result containing the process output and exit code.
+
+Example:
+
+```ts
+import spawn from 'cross-spawn' // [!code --]
+import { x } from 'tinyexec' // [!code ++]
+
+const child = spawn('npm', ['list', '-g']) // [!code --]
+const { stdout, exitCode } = await x('npm', ['list', '-g']) // [!code ++]
+
+child.stdout?.on('data', (output) => console.log(output.toString())) // [!code --]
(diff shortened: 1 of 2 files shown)


gameroman (changes requested):
We should probably add an additional explanation for migrating from it

43081j:
agreed, it should have a `url`, a new doc which just lists tinyexec for now.

may also be worth noting in there that you may not need it at all. e.g. if you're only targeting linux or something

anilloutombam:
Updated the documentation with the `tinyexec` migration and added a note about using `node:child_process` directly when cross-platform handling isn’t required.

docs/modules/cross-spawn.md
```diff
@@ -0,0 +1,33 @@
...
+import { x } from 'tinyexec' // [!code ++]
+
+const child = spawn('npm', ['list', '-g']) // [!code --]
+const { stdout, exitCode } = await x('npm', ['list', '-g']) // [!code ++]
+
+child.stdout?.on('data', (output) => console.log(output.toString())) // [!code --]
+console.log(stdout, exitCode) // [!code ++]
+```
+
+If your application only targets Linux, or otherwise does not require the Windows compatibility handling provided by `cross-spawn`, you may not need another package. You can use the built-in `node:child_process` module directly:
```
43081j: i would drop the example below this, and move this text to above the `tinyexec` header as a notice:

```md
> [!NOTE]
> If your application does not need to support Windows, you can likely drop `cross-spawn` entirely as it solely exists to provide Windows spawn compatibility. You can use the built-in `node:child_process` in that case.
```

anilloutombam: Updated as suggested: moved the native alternative into a notice above the `tinyexec` section and removed the additional example.

PR #907, opened 2026-08-12
feat(native): add tweetnacl to replacements

### LINKED ISSUE #843: [Replacement]: `tweetnacl`
### Package to replace

`tweetnacl`

### Suggested replacement(s)

Native web `crypto`

Before

```js
import nacl from 'tweetnacl'
nacl.sign.detached.verify(message, sig, pubKey)
```

After

```js
const publicKey = await crypto.subtle.importKey(
  'raw',
  pubKey,
  { name: 'Ed25519' },
  false,
  ['verify']
)

const isVerified = await crypto.subtle.verify(
  { name: 'Ed25519' },
  publicKey,
  sig,
  message
)
```

### Manifest type

native (replaceable by a built-in platform feature)

### Rationale

The features are available natively now

### Availability

_No response_

### Code example (optional)

```js

```

dreyfus92:
### 🔗 Linked issue

closes: #843

### 📚 Description

adds `tweetnacl` to replacements

### DIFF
--- a/manifests/native.json
+++ b/manifests/native.json
@@ -1641,15 +1641,6 @@
         "compatKey": "api.btoa"
       }
     },
-    "crypto.getRandomValues": {
-      "id": "crypto.getRandomValues",
-      "type": "native",
-      "url": {"type": "mdn", "id": "Web/API/Crypto/getRandomValues"},
-      "webFeatureId": {
-        "featureId": "web-cryptography",
-        "compatKey": "api.Crypto.getRandomValues"
-      }
-    },
     "crypto.randomBytes": {
       "id": "crypto.randomBytes",
       "type": "native",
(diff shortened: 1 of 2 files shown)


manifests/native.json
```diff
@@ -3268,6 +3313,18 @@
       "moduleName": "trim",
       "replacements": ["String.prototype.trim"]
     },
+    "tweetnacl": {
+      "type": "module",
+      "moduleName": "tweetnacl",
+      "replacements": [
+        "crypto.getRandomValues",
```
43081j: this is interpreted as if the `replacements` are all equivalent. it needs to be `crypto` i guess.. or `crypto.subtle`

gameroman: Yeah, should be just `crypto` I think

dreyfus92: crypto is a wide scope, so i think it would be better to set it to `crypto.subtle`

gameroman: since it includes `crypto.getRandomValues`, it should be `crypto`

43081j: you're both right. `crypto` isn't very useful although it is technically correct.

`crypto.subtle` is more useful since it points at something specific.

since this is a documented replacement though, it doesn't matter too much as we show the link to the e18e doc rather than the module. as long as the engine info ends up right, happy to go with either.

if you two can't decide, i'd use `crypto.subtle` just so its more specific for the engines calculation. but doubt it makes much difference

gameroman (approved):

PR #1008, opened 2026-08-26
chore: add issue validation workflow

43081j:
### 🔗 Linked issue

N/A

### 📚 Description

Adds a basic validation flow for issues.


gameroman:
I'd also change the existing issue template

I think we should split the existing one into 3 and remove the `Manifest type` dropdown 

So that we can have better template for each type of the replacement

43081j:
can you open a separate issue for that?

we can tackle it in its own change 👍

PR #1028, opened 2026-08-27
feat: add istanbul-lib-coverage to preferred replacements

### LINKED ISSUE #993: [Replacement]: `istanbul-lib-coverage`
### Package to replace

`istanbul-lib-coverage`

### Suggested replacement(s)

`@vitest/istanbul-lib-coverage` is a modern, maintained fork of `istanbul-lib-coverage`

### Manifest type

### Rationale

`istanbul-lib-coverage` is unmaintained

### Availability

_No response_

### Code example (optional)

```js

```

ChrisJr404:
Adds `istanbul-lib-coverage` to the preferred manifest, pointing at the maintained `@vitest/istanbul-lib-coverage` fork, since the original is no longer maintained.

The fork keeps the same API so migrating is mostly just swapping the import, and it ships as ESM with its own types. I added a docs page covering that.

Closes #993

### DIFF
--- a/docs/modules/istanbul-lib-coverage.md
+++ b/docs/modules/istanbul-lib-coverage.md
@@ -1,24 +1,18 @@
 ---
-description: Replace the unmaintained istanbul-lib-coverage package with the maintained @vitest/istanbul-lib-coverage fork
+description: Modern alternatives to the istanbul-lib-coverage package
 ---
 
 # Replacements for `istanbul-lib-coverage`
 
-`istanbul-lib-coverage` is no longer maintained. The Vitest team publishes
-[`@vitest/istanbul-lib-coverage`](https://www.npmjs.com/package/@vitest/istanbul-lib-coverage),
-a maintained fork that keeps the same API, so migrating is usually just a matter
-of changing the import.
-
-Compared to the original it is:
-
-- Actively maintained
-- Shipped as ESM instead of CommonJS only
-- Bundled with its own type definitions, so you no longer need `@types/istanbul-lib-coverage`
+`istanbul-lib-coverage` is no longer maintained.
 
 ## `@vitest/istanbul-lib-coverage`
 
-The exported functions and classes keep their existing names, so the only change
-in most projects is the package you import from.
+The Vitest team publishes
+[`@vitest/istanbul-lib-coverage`](https://www.npmjs.com/package/@vitest/istanbul-lib-coverage),
+a maintained fork that keeps the same API. The exported functions and classes
+keep their existing names, so in most projects the only change is the package
+you import from.
 
 ```ts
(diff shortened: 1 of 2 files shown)


AriPerkkio (approved):

gameroman:
Should we also add other `istanbul-*` packages in the same PR?

dreyfus92 (commented):
thanks @ChrisJr404 for putting this together, and @AriPerkkio for the quick review.

one thing I was wondering about while reading the docs page: the fork has `engines: node >=22` and ships ESM-only, which is a bigger jump than "just swap the import" for some projects. would it be worth a line or two mentioning that, so people on older Node or CJS aren't caught off guard?

@gameroman good call on the other `istanbul-*` packages, but i'd rather keep this PR scoped to `istanbul-lib-coverage` since the linked issue covers just this one and it's already approved. would you mind opening a tracking issue for the rest?

docs/modules/istanbul-lib-coverage.md
```diff
@@ -0,0 +1,31 @@
+---
+description: Replace the unmaintained istanbul-lib-coverage package with the maintained @vitest/istanbul-lib-coverage fork
```
43081j: this should probably be worded similar to the others, "Modern alternatives to ..."

docs/modules/istanbul-lib-coverage.md
```diff
@@ -0,0 +1,31 @@
...
+---
+
+# Replacements for `istanbul-lib-coverage`
+
+`istanbul-lib-coverage` is no longer maintained. The Vitest team publishes
```
43081j: the context around the vitest fork should be in the vitest section, leaving only the "no longer maintained" message here i think

ChrisJr404:
Updated the doc per the review:

- Frontmatter now uses the "Modern alternatives to ..." wording like the other pages.
- The intro is just the "no longer maintained" line; the fork context now lives under the `@vitest/istanbul-lib-coverage` section.
- Added a note that the fork is ESM-only and sets `engines.node` to `>=22`, so it is a bigger jump than a plain import swap for CommonJS or older-Node projects (thanks @dreyfus92).

@gameroman I'd like to keep this PR scoped to `istanbul-lib-coverage` since that is what the linked issue covers; happy to see the other `istanbul-*` packages tracked separately.

Heads up on the `validate` check: it fails because `@vitest/istanbul-lib-coverage` is at ~1.9k weekly downloads, under the 10k threshold in `scripts/validate-pr.ts`. That is a policy call on a maintained-but-low-traffic fork rather than something I can change, so I will leave that decision to you all.

43081j (approved):

43081j:
given it is in the vitest org, i'm happy to skip the downloads threshold in this case.

also seems fine to do the other packages in follow ups 👍

PR #1040, opened 2026-08-28
feat: add `escape-html` and `html-escaper` to replacements

### LINKED ISSUE #945: [Replacement]: `escape-html`
### Package to replace

`escape-html`

### Suggested replacement(s)

```js
const map = {
  '&': '&amp;',
  '<': '&lt;',
  '>': '&gt;',
  '"': '&quot;',
  "'": '&#39;'
}

function escapeHtml(string) {
  return string.replace(/[&<>"']/g, char => map[char])
}
```

### Manifest type

### Rationale

It's a simple util 

### Availability

_No response_

### Code example (optional)

```js

```

gameroman:
### 🔗 Linked issue

Closes #945

### 📚 Description

add replacements

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -102,6 +102,12 @@
       "description": "You can use a ternary operator and string concatenation to add a trailing `.0`.",
       "example": "const comverToSemver = (comver) => comver.includes('.') ? `${comver}.0` : `${comver}.0.0`"
     },
+    "snippet::escape-html": {
+      "id": "snippet::escape-html",
+      "type": "simple",
+      "description": "You can use `String.prototype.replace` to escape HTML characters.",
+      "example": "const map = { '&': '&amp;', '<': '&lt;', '>': '&gt;', '\"': '&quot;', \"'\": '&#39;' }\n\nconst escapeHtml = (str) => str.replace(/[&<>\"']/g, c => map[c])"
+    },
     "snippet::for-own": {
       "id": "snippet::for-own",
       "type": "simple",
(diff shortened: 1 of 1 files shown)


dreyfus92:
a few things I'd wonder about before merging:

- **non-string input.** both packages coerce via `String()`, so `escape(42)` returns `"42"`. the snippet throws on non-strings. Would `String(str).replace(...)` be worth adding? template helpers get fed numbers and undefined a lot.

- **`html-escaper` also exports `unescape`.** the snippet only covers `escape`, so anyone using both directions doesn't have a full replacement. should we add an unescape snippet too, or keep this PR to `escape-html` and do `html-escaper` in a follow-up?

on the perf question from james i benchmarked it since it was left open:

| input | escape-html | snippet |
|---|---|---|
| 100 chars, no specials | 85 ns | 167 ns |
| short HTML fragment | 289 ns | 594 ns |
| dense specials | 2.1 µs | 7.4 µs |

so @43081j was right that the loop is there for perf (the repo has a `benchmark/` dir too), but it's ~2x on realistic input, which won't matter outside a hot SSR loop. might be worth one line in the description noting the trade-off so people who picked `escape-html` for throughput aren't surprised.

43081j:
this is non-trivial so i do wonder if we are better using a documented replacement, then give the two options (1. readable but slightly slower, 2. fast but not as readable).

gameroman:
Maybe we should recommend [`fast-escape-html`](https://main.npmx.dev/package/fast-escape-html) instead then?

PR #1049, opened 2026-08-28
feat: add `objectorarray` to replacements

### LINKED ISSUE #1021: [Replacement]: `objectorarray`
### Package to replace

`objectorarray`

### Suggested replacement(s)

```js
const objectorarray = (val) =>
  val != null && typeof val === 'object' && val.constructor !== RegExp;
```

### Manifest type

### Rationale

inlinable

### Availability

_No response_

### Code example (optional)

```js

```

gameroman:
### 🔗 Linked issue

Closes #1021

### 📚 Description

add `objectorarray` to replacements

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -264,6 +264,12 @@
       "description": "You can use `typeof` to check if a value is an object and `Object.getPrototypeOf` to ensure it's a plain object.",
       "example": "const isObject = (obj) => obj && typeof obj === \"object\" && (Object.getPrototypeOf(obj) === null || Object.getPrototypeOf(obj) === Object.prototype);"
     },
+    "snippet::is-object-and-not-regexp": {
+      "id": "snippet::is-object-and-not-regexp",
+      "type": "simple",
+      "description": "You can use `typeof` to check if a value is an object and is not a `RegExp`.",
+      "example": "const isObjectAndNotRegExp = (v) => v != null && typeof v === \"object\" && v.constructor !== RegExp"
+    },
     "snippet::is-object-or-function": {
       "id": "snippet::is-object-or-function",
       "type": "simple",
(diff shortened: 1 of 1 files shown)


43081j:
the package checks if something is an object or an array, but this checks if its an object and not a RegExp.

shouldn't this be `typeof` and `Array.isArray`?

gameroman:
> the package checks if something is an object or an array, but this checks if its an object and not a RegExp.

The package does that https://main.npmx.dev/package-code/objectorarray/v/1.0.5/index.js

43081j:
img

i think maybe the package is just old/terrible? and we should suggest something more sane?

though then it just dumbs down to `typeof val === 'object' && val != null`. no clue why it special cases RegExp but no other class (e.g. Date, Map, etc).

gameroman:
Yeah, I guess it's just because it's old. `Array.isArray` makes more sense now

gameroman:
I think like this

```js
const isPlainObjectOrArray = (val) => {
  if (val === null || typeof val !== 'object') return false
  return Array.isArray(val) || Object.getPrototypeOf(val) === Object.prototype
}
```

43081j:
it isn't actually checking that it is a _plain_ object it seems.

so it may be even simpler...

gameroman:
I think we should keep the snippet as is then, even though the package name is wrong, that's what it does

43081j:
i think it is just poorly named and old, it should be checking if something is an object or an array.

it is trying to do that by checking if it is a non-RegExp object, probably before `Array.isArray` was a thing.

but this seems as simple as `typeof v === 'object' && v != null`. that would match all objects and arrays, which is what the name and description of this package claims it exists for.

gameroman:
The package is used in https://npmx.dev/package/endent made by the same person

All downloads come from there

PR #1088, opened 2026-09-04
feat(native): add is-utf8 to replacements

### LINKED ISSUE #1024: [Replacement]: `is-utf8`
### Package to replace

`is-utf8`

### Suggested replacement(s)

```js
const decoder = new TextDecoder('utf-8', { fatal: true });

const isUtf8 = (bytes) => {
  try {
    decoder.decode(bytes);
    return true;
  } catch {
    return false;
  }
};
```

### Manifest type

### Rationale

inlinable. this variation is even better than the og

### Availability

_No response_

### Code example (optional)

```js

```

dreyfus92:
### 🔗 Linked issue

closes: #1024

### 📚 Description

adds `is-utf8` to the native manifest

### DIFF
--- a/manifests/native.json
+++ b/manifests/native.json
@@ -1719,6 +1719,13 @@
         "compatKey": "api.btoa"
       }
     },
+    "buffer.isUtf8": {
+      "id": "buffer.isUtf8",
+      "type": "native",
+      "description": "Checks whether a buffer, typed array or array buffer contains valid UTF-8, including control characters such as NUL and ESC.",
+      "url": {"type": "node", "id": "api/buffer.html#bufferisutf8input"},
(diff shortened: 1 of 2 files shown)


manifests/native.json
```diff
@@ -1719,6 +1719,13 @@
         "compatKey": "api.btoa"
       }
     },
+    "buffer.isUtf8": {
+      "id": "buffer.isUtf8",
+      "type": "native",
+      "description": "Checks whether a buffer contains valid UTF-8, though unlike `is-utf8` it accepts control characters such as NUL and ESC.",
```
gameroman: `description` of the native feature shouldn't mention packages

that should be in the doc instead

gameroman (changes requested):

43081j (approved):

gameroman (approved):

PR #1091, opened 2026-09-04
feat(micro-utilities): add is-posix-bracket to replacements

### LINKED ISSUE #1018: [Replacement]: `is-posix-bracket`
### Package to replace

`is-posix-bracket`

### Suggested replacement(s)

```js
const isPosixBracket = (str) =>
  typeof str === 'string' && /\[([:.=+])(?:[^\[\]]|)+\1\]/.test(str);
```

### Manifest type

### Rationale

trivially inlinable

### Availability

_No response_

### Code example (optional)

```js

```

dreyfus92:
### 🔗 Linked issue

closes: #1018

### 📚 Description

adds `is-posix-bracket` to the micro-utilities manifest

### DIFF
--- a/manifests/micro-utilities.json
+++ b/manifests/micro-utilities.json
@@ -322,7 +322,7 @@
       "id": "snippet::is-posix-bracket",
       "type": "simple",
       "description": "You can test for a POSIX bracket expression such as `[:alpha:]` using a regular expression.",
-      "example": "const isPosixBracket = (str) =>\n  typeof str === 'string' && /\\[([:.=+])(?:[^\\[\\]]|)+\\1\\]/.test(str);"
+      "example": "const isPosixBracket = (str) => /\\[([:.=+])(?:[^\\[\\]]|)+\\1\\]/.test(str);"
     },
     "snippet::is-primitve": {
       "id": "snippet::is-primitve",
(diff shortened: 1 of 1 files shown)


manifests/micro-utilities.json
```diff
@@ -318,6 +318,12 @@
       "description": "You can check the first eight bytes of a buffer against the PNG file signature to detect if it is a PNG image.",
       "example": "function isPng(buf) {\n  if (!buf || buf.length < 8) return false;\n  return buf[0] === 0x89 && buf[1] === 0x50\n    && buf[2] === 0x4e && buf[3] === 0x47\n    && buf[4] === 0x0d && buf[5] === 0x0a\n    && buf[6] === 0x1a && buf[7] === 0x0a;\n}"
     },
+    "snippet::is-posix-bracket": {
+      "id": "snippet::is-posix-bracket",
+      "type": "simple",
+      "description": "You can test for a POSIX bracket expression such as `[:alpha:]` using a regular expression.",
+      "example": "const isPosixBracket = (str) =>\n  typeof str === 'string' && /\\[([:.=+])(?:[^\\[\\]]|)+\\1\\]/.test(str);"
```
```suggestion
      "example": "const isPosixBracket = (str) => /\\[([:.=+])(?:[^\\[\\]]|)+\\1\\]/.test(str);"
```

gameroman (changes requested):

manifests/micro-utilities.json
dreyfus92: the only reason i kept the guard is that `test` coerces its arg, so without it `['[:alpha:]']` comes back true, and `Symbol()` / `Object.create(null)` throw instead of returning false. the og package returns false for all of those.

not a strong opinion though, happy to drop it if you'd rather keep it to one line. wdyt?

gameroman: since `isPosixBracket` expects already expects a string and no other snippet does that we should remove that

gameroman (approved):

PR #1117, opened 2026-09-13
feat(micro-utilities): add is-upper-case to replacements

### LINKED ISSUE #1004: [Replacement]: `is-upper-case`
### Package to replace

`is-upper-case`

### Suggested replacement(s)

```js
const isUpperCase = (input, locale) =>
  input.toLocaleUpperCase(locale) === input &&
  input !== input.toLocaleLowerCase(locale)
```

### Manifest type

### Rationale

deprecated by the maintainer. we can either use replacement from the deprecation banner or the one above 

### Availability

_No response_

### Code example (optional)

```js

```

kamalesh404:
### 🔗 Linked issue

closes: #1004

### 📚 Description

adds `is-upper-case` to the micro-utilities manifest (snippet + module mapping), following the same shape as the merged `is-lower-case` entry. The implementation is the locale-aware check suggested in the issue for the deprecated `is-upper-case` package: a string is entirely upper case when it equals its `toLocaleUpperCase` form and contains cased characters.


gameroman:
Please fill in the PR template correctly and fix formatting

kamalesh404:
Thanks for the review! I've updated the PR body to follow the template (linked issue + description). Could you point me to which formatting issues you'd like fixed in the manifest?

gameroman:
> Thanks for the review! I've updated the PR body to follow the template (linked issue + description). Could you point me to which formatting issues you'd like fixed in the manifest?

image

kamalesh404:
Thanks for the screenshot — that made it clear. I rebuilt the branch on current main: the whole-file key re-sort is gone (the earlier push had rewritten every entry to a different key order and added a stray `id` field to the mapping). The diff is now a minimal 11-line insertion — the `snippet::is-upper-case` entry with `id/type/description/example` key order and the `is-upper-case` mapping under `mappings` with `type/moduleName/replacements`, both placed alphabetically, matching the merged `is-lower-case` entry. The example now also uses the locale-aware implementation suggested in #1004.

43081j:
Evidently we're talking to an agent here. It is important PRs are submitted by someone who is going to be responsible for the change, not their agent. So can you reply at some point yourself, not via your agent, to ensure this isn't fully automated.

If you're not able to do that because a human isn't in the loop, let us know and we can deal with it differently.

PR #1118, opened 2026-09-14
chore: split issue template per manifest type

### LINKED ISSUE #1014: change existing issue template
I think we should change the existing issue template

I think we should split the existing one into 3 and remove the `Manifest type` dropdown 

So that we can have better template for each type of the replacement

dreyfus92:
### 🔗 Linked issue

closes: #1014

### 📚 Description

splits the replacement issue template into one per manifest (`native`, `micro-utilities`, `preferred`) and drops the `Manifest type` dropdown, each template only asks for what that manifest needs

each template also has a required package link field (npmx or npmjs) so the link comes from whoever opens the issue


.github/workflows/validate-issue.yml
```diff
@@ -35,7 +35,15 @@ jobs:
...
           ISSUE_TITLE: ${{ github.event.issue.title }}
+          ISSUE_BODY: ${{ github.event.issue.body }}
         run: node scripts/validate-issue.ts
+      - name: Add npmx Link
+        if: hashFiles('issue-body.md') != ''
+        env:
+          GH_TOKEN: ${{ github.token }}
+          ISSUE_NUMBER: ${{ github.event.issue.number }}
+          GH_REPO: ${{ github.repository }}
+        run: gh issue edit "$ISSUE_NUMBER" --body-file issue-body.md
```
43081j, line 46: i wouldn't bother with this bit. if its in the template, people should be adding an npmjs or npmx link themselves.

if they don't, they didn't follow the template.

this changes the workflow from being pure validation to something that mutates issues

dreyfus92: sounds good, i will fix this 🫠

43081j (approved):

